PR #1, opened 2015-03-17
Add transform:none to the UA style sheet

foolip:
https://www.w3.org/Bugs/Public/show_bug.cgi?id=27931


annevk:
I'm still not great at merging in a way that credits the original author :-(

https://github.com/whatwg/fullscreen/commit/a7d0ccf09c9ec01f987856dd9bc99e46a3f33fe2

foolip:
Doesn't matter :)

PR #43, opened 2016-07-11
Allow requestFullscreen() to be called from an orientation change event (closes #34)

### LINKED ISSUE #34: Allow fullscreen to be triggered by an orientation change event
On mobile devices, a common user experience is to have videos play inline when the device is in portrait mode and fullscreen when the user changes the orientation to landscape. Unfortunately, this is not possible on the Web because of the user gesture requirement to go fullscreen.

Changing the user gesture definition to include orientation change is unlikely a good idea because it wouldn't be appropriate for more dangerous features than fullscreen (eg. gUM). However, for fullscreen, and maybe other features, orientation change can be a good signal.

Chrome is willing to ship this ASAP. Anne, would Mozilla be interested to implement this change?

/CC @foolip

mounirlamouri:
This is using https://github.com/w3c/screen-orientation/pull/90

### DIFF
--- a/Overview.html
+++ b/Overview.html
@@ -9,7 +9,7 @@
 
 <hgroup>
  <h1 class="p-name">Fullscreen API</h1>
- <h2 class="no-num no-toc">Living Standard — Last Updated 20 July 2016</h2>
+ <h2 class="no-num no-toc">Living Standard — Last Updated 21 July 2016</h2>
 </hgroup>
 
  <dl>
@@ -195,6 +195,15 @@ steps:
 
 <p><dfn id="fullscreen-is-supported">Fullscreen is supported</dfn> if there is no previously-established user
 preference, security risk, or platform limitation.
+
+<p>An algorithm is <dfn id="allowed-to-request-fullscreen">allowed to request fullscreen</dfn> if one of the following is true:
+
+ <ul>
+  <li><p>The algorithm is <a class="external" data-anolis-spec="html" href="https://html.spec.whatwg.org/multipage/interaction.html#triggered-by-user-activation">triggered by user activation</a>.
+
+  <li><p>The algorithm is
+  <a class="external" data-anolis-spec="screen-orientation" href="https://w3c.github.io/screen-orientation/#dfn-triggered-by-a-user-generated-orientation-change">triggered by a user generated orientation change</a>.
+ </ul>
 <!-- cross-process -->
 
 
@@ -281,10 +290,9 @@ method, when invoked, must run these steps:
    <li><p>The <a href="#fullscreen-element-ready-check">fullscreen element ready check</a> for <var title="">pending</var>
    returns false.
 
-   <li><p>This algorithm is not
-   <a class="external" data-anolis-spec="html" href="https://html.spec.whatwg.org/multipage/browsers.html#allowed-to-show-a-popup">allowed to show a popup</a>.
-
    <li><p><a href="#fullscreen-is-supported" title="Fullscreen is supported">Fullscreen is <em>not</em> supported</a>.
+
+   <li><p>This algorithm is not <a href="#allowed-to-request-fullscreen">allowed to request fullscreen</a>.
   </ul>
 
  <li><p>Return <var>promise</var>, and run the remaining steps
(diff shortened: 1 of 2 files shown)


mounirlamouri:
@foolip PTAL

foolip:
@annevk should review this. I see that Fullscreen hasn't adopted the "triggered by user activation" rename yet, with that and a similarly named concept in the Screen Orientation spec this would look pretty clean I think.

I'll keep noting at each step that I don't think it's a good idea to have an API that cannot be implemented without bad frames. Before this is implemented I think that some form of sign-off from the relevant people at each browser engine would be good, and if none think that this matters then I will be quiet. (Shipping it and having them complain about exactly this after the fact would be sad.)

mounirlamouri:
For the record, Mozilla stated publicly that they were okay with the bad frame. I sent an email to Microsoft and Apple (@foolip was CC'd). Microsoft said that they were okay with the feature and with the bad frame. Apple made no comment. All the browsers implementing the Screen Orientation API and the Fullscreen API on mobile are supportive of this change.

mounirlamouri:
@foolip @annevk  PTAL. I simplified the logic living in the fullscreen spec.

annevk:
This looks okay. I'd prefer if this pointed at the new term in HTML and has a reference to Screen Orientation API.

annevk:
@foolip could you point to a canonical explanation of the bad frame problem? Might be useful in the future and we should maybe include it in the commit message.

mounirlamouri:
https://github.com/whatwg/xref/pull/12 is doing the changes required in xref.

mounirlamouri:
@annevk PTAL, your comments are applied. I can update the commit message if @foolip wants.

foolip:
I picked up the term "bad frame" from the [Intent to Deprecate: overflowchanged event](https://groups.google.com/a/chromium.org/d/msg/blink-dev/1RCzsyEyNU8/gs7gLOjIU78J) on blink-dev two years ago, I don't know if this is what it's usually called.

But let me illustrate it with commented code:

```JavaScript
screen.orientation.addEventListener('change', function() {
  // called in the first animation frame task after the user rotates
  // the viewport width and height have changed and there will be a layout before the next paint
  someElement.requestFullscreen();
  document.addEventListener('fullscreenchange', function() {
    // called in the first animation frame task after going fullscreen
    // the viewport size (most likely height) has changed again
    // there will be a layout before the next paint
    // probably some things will move vertically
  });
});

File: Overview.html [hunk 1]
```diff
@@ -757,6 +765,7 @@ <h2 class="no-num" id="acknowledgments">Acknowledgments</h2>
 João Eiras,
 Josh Soref,
 Matt Falkenhagen,
+Mounir Lamouri,
```
annevk: Mo > Mi.

Overview.src.html
```diff
@@ -180,6 +180,15 @@ <h2 class="no-num no-toc">Table of Contents</h2>
...
 <p><dfn>Fullscreen is supported</dfn> if there is no previously-established user
 preference, security risk, or platform limitation.
+
+<p>An algorithm is <dfn>allowed to request fullscreen</dfn> if one of the following conditions are
+true:
+
+ <ul>
+  <li>The algorithm is <span data-anolis-spec=html>triggered by user activation</span>.
+
+  <li>The algorithm is <span data-anolis-spec=screen-orientation>triggered by a user generated orientation change</span>.
```
annevk: Newline before `<span>` and please use `<li><p>` here too.

Overview.src.html
```diff
@@ -180,6 +180,15 @@ <h2 class="no-num no-toc">Table of Contents</h2>
 
 <p><dfn>Fullscreen is supported</dfn> if there is no previously-established user
 preference, security risk, or platform limitation.
+
+<p>An algorithm is <dfn>allowed to request fullscreen</dfn> if one of the following conditions are
+true:
```
annevk: "if one of the following is true"

Also this might need to be wrapped per 100column rule.

Overview.html
mounirlamouri: One day I will know my alphabet :)

Overview.src.html
mounirlamouri: Done.

Overview.src.html
mounirlamouri: Done

mounirlamouri:
@annevk PTAL. I've applied all the comments.

Issue #47, opened 2016-09-13
Convert to Bikeshed


foolip:
Anolis is no longer maintained, and makes it harder (for me) to test PRs before sending them.

This will not be a manual conversion, but rather the minimum manual changes possible combined with a script to do most of the work, followed by verification of the generated output.

foolip:
Now done. If anyone wants to repeat this with another anolis spec, see https://github.com/whatwg/fullscreen/pull/49/commits for the individual steps.

PR #48, opened 2016-09-13
Prepare for conversion to Bikeshed

foolip:
These are changes that make (some) sense even with anolis, done first
to make the Bikeshed changes smaller:
 * Add ids to headings
 * Don't quote attributes without whitespace
 * &oslash; → ø

Part of #47.

### DIFF
--- a/Overview.html
+++ b/Overview.html
@@ -9,7 +9,7 @@
 
 <hgroup>
  <h1 class="p-name">Fullscreen API</h1>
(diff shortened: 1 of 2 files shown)


foolip:
@annevk, can you generate Overview.html and see that this didn't break things? The difficulty of doing so myself is part of the motivation here :)

annevk:
Do you want me to land this or are there more changes you'd like to make?

foolip:
Please land, the real PR will build on this.

PR #49, opened 2016-09-13
Convert to Bikeshed

foolip:
Optimized for review, when the sum of changes looks good I'll turn it into a single commit.


foolip:
Oh right, this doesn't rewrap anything, I think it makes sense to do that at the same time, WDYT?

fullscreen.bs
```diff
@@ -1,3 +1,60 @@
...
+!Participate: <a href=https://wiki.whatwg.org/wiki/IRC>IRC: #whatwg on Freenode</a>
+!Commits: <a href=https://github.com/whatwg/fullscreen/commits>GitHub whatwg/fullscreen/commits</a>
+!Commits: <a href=https://twitter.com/fullscreenapi>@fullscreenapi</a>
+</pre>
+
+<script src=https://resources.whatwg.org/file-issue.js async></script>
+<script id=head src=https://resources.whatwg.org/dfn.js defer></script>
+
+<pre class=anchors>
+urlPrefix: https://dom.spec.whatwg.org/
```
annevk: Since DOM is in bikeshed, we should not need this I think.

fullscreen.bs
```diff
@@ -1,3 +1,60 @@
...
+        text: concept node document
+        text: concept node remove ext
+        text: concept tree ancestor
+        text: concept tree inclusive ancestor
+        text: concept tree inclusive descendant
+        text: concept tree order
+        text: concept tree root
+    type: interface
+        text: Document
+urlPrefix: https://html.spec.whatwg.org/multipage/
```
annevk: The problem here I suppose is that HTML doesn't export these terms?

annevk:
Rewrap would be nice.

Also, an initial commit that does the renames of Overview.src.html to fullscreen.bs and Overview.html to fullscreen.html would be good (and does no other changes). That way there's less to remove and more git blame.

foolip:
I've rewritten the history to do the rename first, net changes are the same.

fullscreen.bs
foolip: It looks like "browsing context" can be dropped from this list, but Bikeshed doesn't know about "ancestor browsing context", "browsing context container" or "unloading document cleanup steps".

fullscreen.bs
foolip: It looks some things can be used without this block, like "node document". I'll trim it down to just the ones that don't work without this.

foolip:
Ready for proper review now.

fullscreen.bs
```diff
@@ -534,18 +582,18 @@ user.
  <code>top</code> is zero.
 </ul>
 
-<p>To <dfn>add</dfn> an <var>element</var> to a
+<p>To <dfn id=concept-top-layer-add>add</dfn> an <var>element</var> to a
```
annevk: These should probably have `for="top layer"` too?

fullscreen.bs
```diff
@@ -534,18 +582,18 @@ user.
...
-<p>To <dfn>remove</dfn> an <var>element</var> from a
+<p>To <dfn id=concept-top-layer-remove>remove</dfn> an <var>element</var> from a
 <var>top layer</var>, remove <var>element</var> from <var>top layer</var>.
 
 
 <h3 id=::backdrop-pseudo-element><code>::backdrop</code> pseudo-element</h3>
 
 <p>Each element in a <a>top layer</a> has a
-<dfn lt="css pe backdrop"><code>::backdrop</code></dfn> pseudo-element.
+<dfn id=css-pe-backdrop><code>::backdrop</code></dfn> pseudo-element.
```
annevk: Should this have type=selector or does bikeshed handle that based on syntax?

fullscreen.bs
```diff
@@ -564,13 +612,13 @@ pseudo-element either.
 
 <h3 id=:fullscreen-pseudo-class><code>:fullscreen</code> pseudo-class</h3>
 
-<p>The <dfn lt="css pc fullscreen"><code>:fullscreen</code></dfn>
+<p>The <dfn id=css-pc-fullscreen><code>:fullscreen</code></dfn>
```
annevk: Same here.

annevk:
We should probably also add `export` to various `<dfn>` elements?

fullscreen.bs
foolip: That's not enough to remove the need for the id, and also isn't needed to disambiguate. Is it so that HTML could link to this if it used Bikeshed? No other specs use top layer.

fullscreen.bs
foolip: Adding it only adds `type="selector"` to the output, `data-dfn-type="selector"` is already generated automatically.

fullscreen.bs
foolip: Tested, also not needed.

foolip:
I'll export the terms used by HTML, are there any others you think are needed?

foolip:
Well, HTML says "When an element is added to the pending dialog stack, it must also be added to the top layer. When an element is removed from the pending dialog stack, it must be removed from the top layer." without using Fullscreen's (new) add/remove definitions.

How about exporting only add/remove and treating top layer itself as an internal concept?

foolip:
Strike that, the concept itself needs to be exposed in order to say "document's top layer".

foolip:
Comments addressed, added for="top layer" too just in case.

foolip:
Now ready to commit, as a single commit with this message:
```
Editorial: convert to bikeshed

Steps to reproduce:

 * Convert fullscreen.bs by pasting into this one-off tool:
   https://gist.github.com/foolip/f60a1614fb5d194aad4dbff89c54ceb3
 * Manually finish conversion to Bikeshed, tweaking as needed to match
   the existing output as closely as possible.
 * Export top layer and its add/remove concepts for HTML.
 * Rewrap to 100 columns, taking care not to wrap inside elements.
 * Update README.md and remove Makefile.

(See #49 for branch with separate commits.)

Finally, compare the new fullscreen.html with the old:

  for (var e of document.querySelectorAll('*')) {
    if (e.id)
      console.log('id: ' + e.id);
    if (e.hasAttribute('href') && e.className != 'self-link')
      console.log('href: ' + e.getAttribute('href'));
  }

Existing IDs/links affected:

 * #table-of-contents → #toc
 * #refsCSS → #biblio-css (and similar)
 * #anolis-references is gone

The old and new fullscreen.html were also copied from a browser into
plaintext and compared to verify no accidental differences. The biggest
differences are in the references section.

Fixes #49.
```

PR #51, opened 2016-09-14
Editorial: shorten the top layer add/remove IDs

foolip:
Shorten #concept-top-layer-add/remove → #top-layer-add/remove.

These were introduced recently in 766dc872bcf26c012549f7bdfde21c46bebba40c
and are not yet used by HTML. This makes the IDs nicer in preparation
for using them from HTML.


annevk:
This would break existing links to these concepts. I don't think we should be doing that.

foolip:
It's incredibly unlikely that there are any existing links from outside the spec itself. (I thought that Bikeshed had a feature for adding extra IDs, but I can't find it.)

annevk:
Why? Impl routinely link spec stuff.

annevk:
I don't really see the benefit either.

foolip:
Neither Chromium, Gecko or WebKit sources have the string "concept-top-layer" anywhere, and Edge is waiting for Chrome to ship unprefixed Fullscreen, so it's unlikely they've added a comment either.

It seems like an accident that it's #top-layer but #concept-top-layer-add/remove, and an accident that can trivially be undone since it was so recent. I didn't expect any resistance whatsoever :)

annevk:
Well you know, I try to still take https://www.w3.org/Provider/Style/URI.html seriously. I guess if you really want this you can go ahead and do it, but I'd rather close this.

foolip:
This is a bit silly, but I did an HTTP Archive search and also found no hits, and I selfishly want nice URLs to copy into Blink, so I'll go ahead.

PR #52, opened 2016-09-14
Remove redundant recursive step of fullscreen element ready check

### LINKED ISSUE #50: Final recursive step of fullscreen element ready check is redundant
Currently the [fullscreen element ready check](https://fullscreen.spec.whatwg.org/#fullscreen-element-ready-check) is:
 * element's namespace is the HTML namespace or element is an SVG svg or MathML math element.
 * element is [in a document](https://dom.spec.whatwg.org/#in-a-document).
 * element's node document is [allowed to use](https://html.spec.whatwg.org/multipage/embedded-content.html#allowed-to-use) the feature indicated by attribute name allowfullscreen.
 * element's node document's browsing context either has a browsing context container and the fullscreen element ready check returns true for element's node document's browsing context's browsing context container, or it has no browsing context container.

The final point is redundant, as it's guaranteed to be true by the previous two. The "allowed to use check" guarantees that if this isn't the top-level browsing context, then it's connected to it via iframes, which must be in documents, as removing an iframe discards its browsing context.

Will PR.

foolip:
HTML's "allowed to use" only returns true if the document is in the
top-level browsing context, or if it's connected to it via any number of
iframes, all of which must be in a document to have nested browsing
contexts at all, so this final condition would always be true.

Fixes #50.


annevk:
LGTM % update date before landing.

foolip:
Thanks for review, date changed in 9a75db664018c6d79385fa27cdae4d7b3d5fff17, merging.

PR #54, opened 2016-09-15
Update fullscreen spec to be Shadow DOM compliant

### LINKED ISSUE #42: Allow element in Shadow DOM to become fullscreen
As discussed in https://github.com/w3c/webcomponents/issues/180.

See https://github.com/w3c/webcomponents/issues/180#issuecomment-224244250 for the rough consensus there.

TakayoshiKochi:
Spec discussion at https://github.com/w3c/webcomponents/issues/180.

The rough consensus was:
https://github.com/w3c/webcomponents/issues/180#issuecomment-224244250

This resolves #42

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -110,8 +110,8 @@ and <a>add</a> it to its <a>node document</a>'s <a>top layer</a>.
 <p>Whenever the <a>removing steps</a> run with an <var>oldNode</var>, run these steps:
 
 <ol>
- <li><p>Let <var>nodes</var> be <var>oldNode</var>'s <a>inclusive descendants</a> that have their
- <a>fullscreen flag</a> set, in <a>tree order</a>.
+ <li><p>Let <var>nodes</var> be <var>oldNode</var>'s <a>shadow-including inclusive descendants</a>
+ that have their <a>fullscreen flag</a> set, in <a>shadow-including tree order</a>.
 
  <li>
   <p>For each <var>node</var> in <var>nodes</var>, run these substeps:
@@ -153,13 +153,16 @@ partial interface Element {
 
 partial interface Document {
   [LenientSetter] readonly attribute boolean fullscreenEnabled;
-  [LenientSetter] readonly attribute Element? fullscreenElement;
   [LenientSetter] readonly attribute boolean fullscreen; // historical
 
   Promise&lt;void> exitFullscreen();
 
   attribute EventHandler onfullscreenchange;
   attribute EventHandler onfullscreenerror;
+};
+
+partial interface DocumentOrShadowRoot {
+  [LenientSetter] readonly attribute Element? fullscreenElement;
 };
 </pre>
 
@@ -171,20 +174,22 @@ partial interface Document {
  <dd><p>Returns true if <var>document</var> has the ability to display <a>elements</a> fullscreen
  and <a>fullscreen is supported</a>, or false otherwise.
 
- <dt><code><var>document</var> . {{Document/fullscreenElement}}</code>
- <dd><p>Returns <var>document</var>'s <a>fullscreen element</a>.
-
  <dt><code><var>promise</var> = <var>document</var> . {{Document/exitFullscreen()}}</code>
  <dd><p>Stops <var>document</var>'s <a>fullscreen element</a> from being displayed fullscreen and
  fulfills <var>promise</var> when done.
+
+ <dt><code><var>document</var> . {{DocumentOrShadowRoot/fullscreenElement}}</code>
+ <dd><p>Returns <var>document</var>'s <a>fullscreen element</a>.
+
+ <dt><code><var>shadowroot</var> . {{DocumentOrShadowRoot/fullscreenElement}}</code>
+ <dd><p>Returns <var>shadowroot</var>'s <a>fullscreen element</a>.
 </dl>
 
 <p>A <dfn>fullscreen element ready check</dfn> for an <a>element</a> <var>element</var> returns true
 if all of the following are true, and false otherwise:
 
 <ul>
-
- <li><p><var>element</var> is <a>in a document</a>.
+ <li><p><var>element</var> is <a>connected</a>.
 
  <li><p><var>element</var>'s <a>node document</a> is <a>allowed to use</a> the feature indicated by
  attribute name <code>allowfullscreen</code>.
@@ -283,14 +288,27 @@ these steps:
 return true if the <a>context object</a> is <a>allowed to use</a> the feature indicated by attribute
 name <code>allowfullscreen</code> and <a>fullscreen is supported</a>, and false otherwise.
 
-<p>The <dfn attribute for=Document><code>fullscreenElement</code></dfn> attribute's getter must
-return <a>context object</a>'s <a>fullscreen element</a>.
-
 <p>The <dfn attribute for=Document><code>fullscreen</code></dfn> attribute's getter must return
 false if <a>context object</a>'s <a>fullscreen element</a> is null, and true otherwise.
 
-<p class=note>Use
-<a attribute for=Document lt=fullscreenElement><code>document.fullscreenElement</code></a> instead.
+<p class=note>Use the {{DocumentOrShadowRoot/fullscreenElement}} attribute instead.
+
+<p>The
+<dfn attribute for=DocumentOrShadowRoot id=dom-document-fullscreenelement><code>fullscreenElement</code></dfn>
+attribute's getter must run these steps:
+
+<ol>
+ <li><p>If the <a>context object</a> is a <a for=/>shadow root</a> and its
+ <a for=DocumentFragment>host</a> is not <a>connected</a>, then return null.</li>
+
+ <li><p>Let <var>candidate</var> be the result of <a>retargeting</a> <a>fullscreen element</a>
+ against the <a>context object</a>.
+
+ <li><p>If <var>candidate</var> and the <a>context object</a> are in the same <a>tree</a>, then
+ return <var>candidate</var>.
+
+ <li><p>Return null.
+</ol>
 
 <p>To <dfn>collect documents to unfullscreen</dfn> given <var>doc</var>, run these steps:
 
@@ -461,8 +479,8 @@ following characteristics:
  is the viewport, and the initial containing block otherwise.
 
  <li><p>If it is an element, it and its <a><code>::backdrop</code></a> pseudo-element are not
- rendered if its <a>inclusive ancestor</a> has the <code>display</code> property set to
- <code>none</code>.
+ rendered if its <a>shadow-including inclusive ancestor</a> has the <code>display</code> property
+ set to <code>none</code>.
 
  <li><p>If its specified <code>display</code> property is <code>contents</code>, it computes to
  <code>block</code>.
@@ -503,11 +521,17 @@ properties apply to this pseudo-element either.
 <h3 id=:fullscreen-pseudo-class><code>:fullscreen</code> pseudo-class</h3>
 
 <p>The <dfn id=css-pc-fullscreen><code>:fullscreen</code></dfn> pseudo-class must match any
-<a>element</a> that has its <a>fullscreen flag</a> set.
+<a>element</a> <var>element</var> for which one of the following conditions is true:
 
-<p class="note no-backref">This makes it different from the {{Document/fullscreenElement}} API,
-which returns a <a>document</a>'s <a>fullscreen element</a>.
+<ul>
+ <li><p><var>element</var>'s <a>fullscreen flag</a> is set.
 
+ <li><p><var>element</var> is a <a>shadow host</a> and the result of <a>retargeting</a> its
+ <a>node document</a>'s <a>fullscreen element</a> against <var>element</var> is <var>element</var>.
+</ul>
+
+<p class="note no-backref">This makes it different from the
+{{DocumentOrShadowRoot/fullscreenElement}} API, which returns the topmost <a>fullscreen element</a>.
 
 <h3 id=user-agent-level-style-sheet-defaults>User-agent level style sheet defaults</h3>
 <!-- HTML's "The CSS user agent style sheet and presentational hints" section uses this term -->
--- a/fullscreen.html
+++ b/fullscreen.html
@@ -71,7 +71,7 @@ pre.idl.highlight { color: #708090; }
    <p data-fill-with="logo"><a class="logo" href="https://whatwg.org/"> <img alt="WHATWG" height="100" src="https://resources.whatwg.org/logo-fullscreen.svg"> </a> </p>
    <hgroup>
     <h1 class="p-name no-ref" id="title">Fullscreen API</h1>
-    <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Living Standard — Last Updated <time class="dt-updated" datetime="2016-09-27">27 September 2016</time></span></h2>
+    <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Living Standard — Last Updated <time class="dt-updated" datetime="2016-09-30">30 September 2016</time></span></h2>
    </hgroup>
    <div data-fill-with="spec-metadata">
     <dl>
@@ -152,7 +152,7 @@ stated otherwise it is unset. </p>
    <p>Whenever the <a data-link-type="dfn" href="https://dom.spec.whatwg.org/#concept-node-remove-ext">removing steps</a> run with an <var>oldNode</var>, run these steps: </p>
    <ol>
     <li>
-     <p>Let <var>nodes</var> be <var>oldNode</var>’s <a data-link-type="dfn" href="https://dom.spec.whatwg.org/#concept-tree-inclusive-descendant">inclusive descendants</a> that have their <a data-link-type="dfn" href="#fullscreen-flag">fullscreen flag</a> set, in <a data-link-type="dfn" href="https://dom.spec.whatwg.org/#concept-tree-order">tree order</a>. </p>
+     <p>Let <var>nodes</var> be <var>oldNode</var>’s <a data-link-type="dfn" href="https://dom.spec.whatwg.org/#concept-shadow-including-inclusive-descendant">shadow-including inclusive descendants</a> that have their <a data-link-type="dfn" href="#fullscreen-flag">fullscreen flag</a> set, in <a data-link-type="dfn" href="https://dom.spec.whatwg.org/#concept-shadow-including-tree-order">shadow-including tree order</a>. </p>
     <li>
      <p>For each <var>node</var> in <var>nodes</var>, run these substeps: </p>
      <ol>
@@ -180,13 +180,16 @@ security risk, or platform limitation. </p>
 
 <span class="kt">partial</span> <span class="kt">interface</span> <a class="nv idl-code" data-link-type="interface" href="https://dom.spec.whatwg.org/#document">Document</a> {
   [<a class="nv idl-code" data-link-type="extended-attribute" href="https://heycam.github.io/webidl/#LenientSetter">LenientSetter</a>] <span class="kt">readonly</span> <span class="kt">attribute</span> <span class="kt">boolean</span> <a class="nv idl-code" data-link-type="attribute" data-readonly="" data-type="boolean" href="#dom-document-fullscreenenabled">fullscreenEnabled</a>;
-  [<a class="nv idl-code" data-link-type="extended-attribute" href="https://heycam.github.io/webidl/#LenientSetter">LenientSetter</a>] <span class="kt">readonly</span> <span class="kt">attribute</span> <a class="n" data-link-type="idl-name" href="https://dom.spec.whatwg.org/#element">Element</a>? <a class="nv idl-code" data-link-type="attribute" data-readonly="" data-type="Element?" href="#dom-document-fullscreenelement">fullscreenElement</a>;
   [<a class="nv idl-code" data-link-type="extended-attribute" href="https://heycam.github.io/webidl/#LenientSetter">LenientSetter</a>] <span class="kt">readonly</span> <span class="kt">attribute</span> <span class="kt">boolean</span> <a class="nv idl-code" data-link-type="attribute" data-readonly="" data-type="boolean" href="#dom-document-fullscreen">fullscreen</a>; // historical
 
   <span class="kt">Promise</span>&lt;<span class="kt">void</span>> <a class="nv idl-code" data-link-type="method" href="#dom-document-exitfullscreen">exitFullscreen</a>();
 
   <span class="kt">attribute</span> <a class="n" data-link-type="idl-name" href="https://html.spec.whatwg.org/multipage/webappapis.html#eventhandler">EventHandler</a> <a class="nv idl-code" data-link-type="attribute" data-type="EventHandler" href="#handler-document-onfullscreenchange">onfullscreenchange</a>;
   <span class="kt">attribute</span> <a class="n" data-link-type="idl-name" href="https://html.spec.whatwg.org/multipage/webappapis.html#eventhandler">EventHandler</a> <a class="nv idl-code" data-link-type="attribute" data-type="EventHandler" href="#handler-document-onfullscreenerror">onfullscreenerror</a>;
+};
+
+<span class="kt">partial</span> <span class="kt">interface</span> <a class="nv idl-code" data-link-type="interface" href="https://dom.spec.whatwg.org/#documentorshadowroot">DocumentOrShadowRoot</a> {
+  [<a class="nv idl-code" data-link-type="extended-attribute" href="https://heycam.github.io/webidl/#LenientSetter">LenientSetter</a>] <span class="kt">readonly</span> <span class="kt">attribute</span> <a class="n" data-link-type="idl-name" href="https://dom.spec.whatwg.org/#element">Element</a>? <a class="nv idl-code" data-link-type="attribute" data-readonly="" data-type="Element?" href="#dom-document-fullscreenelement">fullscreenElement</a>;
 };
 </pre>
    <dl class="domintro">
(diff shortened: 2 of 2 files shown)


TakayoshiKochi:
@annevk @foolip  Could you take a look?
@hayatoito FYI

The changes are mainly 3 points:
- Replace "document" with "document tree" where necessary, to allow elements in shadow can become fullscreen
- To match `:fullscreen` against a shadow host which contain a fullscreen element in its shadow, I added sentences to set/unset "fullscreen flag" on it.
- Move `Document.fullscreenElement` to `DocumentOrShadowRoot.fullscreenElement`

fullscreen.bs
```diff
@@ -78,25 +78,29 @@ unset.
...
-<p>To <dfn>fullscreen an <var>element</var></dfn> within a <var>document</var>, set the
-<var>element</var>'s <a>fullscreen flag</a> and <a>add</a> it to <var>document</var>'s
-<a>top layer</a>.
-
-<p>To <dfn>unfullscreen an <var>element</var></dfn> within a <var>document</var>, unset the
-<var>element</var>'s <a>fullscreen flag</a> and <a>iframe fullscreen flag</a> (if any), and
-<a>remove</a> it from <var>document</var>'s <a>top layer</a>.
-
-<p>To <dfn>unfullscreen a <var>document</var></dfn>,
+<p>All <a>document trees</a> have an associated <dfn>fullscreen element</dfn>. The
```
annevk: This should still say "document".

fullscreen.bs
```diff
@@ -78,25 +78,29 @@ unset.
...
+<a>fullscreen element</a> is the topmost <a>element</a> in the <a>document tree</a>'s
```
annevk: This also doesn't quite make sense, especially since you seem to want to include shadow trees.

fullscreen.bs
```diff
@@ -191,7 +203,7 @@ if all of the following are true, and false otherwise:
  <a href=https://www.w3.org/Math/draft-spec/chapter2.html#interf.toplevel>MathML <code>math</code></a>
  element. [[!SVG]] [[!MATHML]]
 
- <li><p><var>element</var> is <a>in a document</a>.
+ <li><p><var>element</var> is <a>in a document tree</a>.
```
annevk: This excludes shadow trees, seems wrong?

fullscreen.bs
foolip: Where should the term "document tree" be used? From DOM it looks like a document tree is a set of objects and not a concrete object, so one can't really associate things with them?

(Also, top layer itself will remain a per-document concept, and "fullscreen element" is just a shorthand for an element within, so I think that it was already odd to say that a document has an associated fullscreen element, rather it's just a helper concept.)

fullscreen.bs
```diff
@@ -289,14 +301,17 @@ these steps:
...
 <p>The <dfn attribute for=Document><code>fullscreen</code></dfn> attribute's getter must return
 false if <a>context object</a>'s <a>fullscreen element</a> is null, and true otherwise.
 
 <p class=note>Use
-<a attribute for=Document lt=fullscreenElement><code>document.fullscreenElement</code></a> instead.
+<a attribute for=DocumentOrShadowRoot lt=fullscreenElement><code>document.fullscreenElement</code></a>
+instead.
+
+<p>The <dfn attribute for=DocumentOrShadowRoot><code>fullscreenElement</code></dfn> attribute's
+getter must return the result of <a>retargeting</a> <a>context object</a>'s
```
foolip: It's unfortunate to have divergence between the definition of "fullscreen element" and `fullscreenElement`. Can "fullscreen element" instead be defined for documents and shadow roots so that it can be used directly here?

fullscreen.bs
```diff
@@ -78,25 +78,29 @@ unset.
...
-
-<p>To <dfn>unfullscreen a <var>document</var></dfn>,
+<p>All <a>document trees</a> have an associated <dfn>fullscreen element</dfn>. The
+<a>fullscreen element</a> is the topmost <a>element</a> in the <a>document tree</a>'s
+<a>top layer</a> whose <a>fullscreen flag</a> is set, if any, and null otherwise.
+
+<p>To <dfn>fullscreen an <var>element</var></dfn> within a <a>document tree</a> whose root is
+<var>document</var>, set the <var>element</var>'s <a>fullscreen flag</a> and <a>add</a> it to
+<var>document</var>'s <a>top layer</a>. If the <var>element</var> is in a <a>shadow tree</a>,
+set <a>fullscreen flags</a> of all <a>shadow hosts</a> in its <a>shadow-including ancestors</a>.
```
foolip: I wonder if, similar to the "iframe fullscreen flag", the shadow hosts will need a flag to say when it was fullscreen'd explicitly and not?

hayatoito: "all shadow hosts in its shadow-including ancestor" would pick-up *false-positive* shadow-hosts.

For example, we don't want to pick up all my-element1, my-element2 and my-element3.

```
<my-element1>
  <my-element2>
    <my-element3>
    </my-element3>
  </my-element2>
</my-element1>

Comment (hayatoito) [inline fullscreen.bs]: I think a better ways is:

- Don't attempt to set flag of multiple elements. Only one element of the flag should be set. Keep the status quo.
- Rather, update the definition of `:fullscreen`.

That would make the spec simpler.

File: fullscreen.bs (hunk 1 above)
Comment (TakayoshiKochi) [inline fullscreen.bs]: FYI "document tree" is defined at [DOM spec](https://dom.spec.whatwg.org/#concept-document-tree).

I seem to misunderstand 'document tree', actually it is restricted to the node tree whose root is document, and excludes any shadow trees.
I thought it was a replacement for 'shadow-including document'.

I'm rewriting the patch.

File: fullscreen.bs (hunk 3 above)
Comment (TakayoshiKochi) [inline fullscreen.bs]: I was wrong, this should have been "connected".

File: fullscreen.bs (hunk 4 above)
Comment (TakayoshiKochi) [inline fullscreen.bs]: I followed [activeElement](https://w3c.github.io/webcomponents/spec/shadow/#active-element) and [pointerLockElement](https://w3c.github.io/pointerlock/#widl-DocumentOrShadowRoot-pointerLockElement).

I prefer the current way, the "fullscreen element" is the real full screen element, and this API returns adjusted element for the context object.

File: fullscreen.bs [hunk 6]
```diff
@@ -78,25 +78,29 @@ unset.
...
-<p>To <dfn>unfullscreen an <var>element</var></dfn> within a <var>document</var>, unset the
-<var>element</var>'s <a>fullscreen flag</a> and <a>iframe fullscreen flag</a> (if any), and
-<a>remove</a> it from <var>document</var>'s <a>top layer</a>.
-
-<p>To <dfn>unfullscreen a <var>document</var></dfn>,
+<p>All <a>document trees</a> have an associated <dfn>fullscreen element</dfn>. The
+<a>fullscreen element</a> is the topmost <a>element</a> in the <a>document tree</a>'s
+<a>top layer</a> whose <a>fullscreen flag</a> is set, if any, and null otherwise.
+
+<p>To <dfn>fullscreen an <var>element</var></dfn> within a <a>document tree</a> whose root is
```
annevk: That seems like a good change to make, yes.

fullscreen.bs
TakayoshiKochi: Oh, I see now why "iframe fullscreen flag" exists.

I'll give a thought (to myself) on how `:fullscreen` should match when in the following HTML
either `<video>` or `<div class="video-content">` become a fullscreen element.
```html
<video>
  <#shadow-root>
    <div class="video-frame">
       <div class="video-content"></div>
       <div class="playback-control"></div>
    </div>
  </#shadow-root>
</video>
```

fullscreen.bs
TakayoshiKochi: Oh, but it might be easier to define `:fullscreen` if we define `fullscreenElement` as you suggested. Let me work on it.

fullscreen.bs
TakayoshiKochi: Created a separate PR for this.
https://github.com/whatwg/fullscreen/pull/55

foolip: Thanks!

fullscreen.bs
foolip: That particular case doesn't actually work because `attachShadow({ mode: 'open' /* or 'closed' */ })` throws an exception, but yeah, the general question is what should happen you first fullscreen the shadow host and an element in its shadow tree, is that different from directly fullscreen the element in the shadow tree?

fullscreen.bs
foolip: You're probably right that having the concept return the topmost element with the flag set in the top layer makes things easier.

I think there's no disagreement about what `fullscreenElement` should return, so the question is what should happen to the other places that use [fullscreen element](https://fullscreen.spec.whatwg.org/#fullscreen-element). Those are:
 * [fully exit fullscreen](https://fullscreen.spec.whatwg.org/#fully-exit-fullscreen), which I *think* can continue to be a document-only concept.
 * The removing steps, where I think having the "unfullscreen element" steps traverse all shadow trees is actually redundant, because the removing steps will be run for those nodes too?
 * In [`requestFullscreen()`](https://fullscreen.spec.whatwg.org/#dom-element-requestfullscreen), where it's just used to avoid an unnecessary event.
 * In [exit fullscreen](https://fullscreen.spec.whatwg.org/#exit-fullscreen), where some tweak are probably needed if you make exit fullscreen per-DocumentOrShadowRoot instead of per-Document.

I'll wait to see the next iteration, thanks for working through this!

fullscreen.bs
foolip: It would be odd I think if `document.fullscreenElement` is a shadow host with the "top" fullscreen element inside, if that element doesn't match the `:fullscreen` selector like an `iframe` would in the same scenario.

That `:fullscreen` matches any element with the fullscreen flag set is a recent change: 4a2644a888bf04bcf9286f969bc750bcd16ff349. On the benefits, it should avoid unnecessary style changes to elements which (usually) aren't visible, and it makes the selector matching trivial.

Still, I'm pretty sure that just walking the ancestor shadow hosts and setting/unsetting their fullscreen flags when fullscreening/unfullscreening an element isn't workable. Take this case:
```
<top-level-host>
 <#shadow-root>
  <inner-host-1>
   <#shadow-root>
    <fullscreen-element-1/>
   </#shadow-root>
  </inner-host-1>
  <inner-host-2>
   <#shadow-root>
    <fullscreen-element-2/>
   </#shadow-root>
  </inner-host-2>
 </#shadow-root>
</top-level-host>
```

If `fullscreen-element-1` goes fullscreen, then `fullscreen-element-2` and then `fullscreen-element-1` is removed and thus unfullscreened, then `top-level-host` mustn't have its flag unset.

fullscreen.bs
TakayoshiKochi: Thanks for the comment.
As you suggested, if I make changes to the "fullscreen element" definition, it will change
other places as well and it makes the spec more complicated.
So I'd stay with the current definition of "fullscreen element" and make adjustments at
`DocumentOrShadowRoot.fullscreenElement` API.

TakayoshiKochi:
Updated the change for addressing comments.
Now we don't set fullscreen flag on ancestor shadow hosts, instead :fullscreen definition is updated.

Could you take another look?

fullscreen.bs
```diff
@@ -505,9 +516,16 @@ properties apply to this pseudo-element either.
 <p>The <dfn id=css-pc-fullscreen><code>:fullscreen</code></dfn> pseudo-class must match any
 <a>element</a> that has its <a>fullscreen flag</a> set.
 
-<p class="note no-backref">This makes it different from the {{Document/fullscreenElement}} API,
+<p class="note no-backref">This makes it different from the
+{{DocumentOrShadowRoot/fullscreenElement}} API,
 which returns a <a>document</a>'s <a>fullscreen element</a>.
```
foolip: In order to avoid repeating the details about "or shadow root" and retargeting here, mayb just say "This makes it different from the fullscreenElement API, which returns the topmost fullscreen element", without linking to the definition of "fullscreen element"? I fear that trying to make this technically correct will just make it harder to understand what the difference is.

fullscreen.bs
```diff
@@ -505,9 +516,16 @@ properties apply to this pseudo-element either.
...
 
+For each <a>node tree</a>, if the result of <a>retargeting</a> its <a>shadow-including root</a>'s
```
foolip: This is about making the shadow host also match the `:fullscreen` selector, right? The shadow host isn't being automatically added to the top layer, so it won't have a `::backdrop` pseudo-element and would never be visible unless the fullscreen element has a transparent background.

In this revision of this PR, shadow trees are handled quite unlike iframes, and I think that actually turned out better, so can we take it even further?

(Not sure how to implement this otherwise, actually doing this as part of selector matching would be too slow.)

fullscreen.bs
```diff
@@ -171,12 +174,17 @@ partial interface Document {
...
- <dt><code><var>document</var> . {{Document/fullscreenElement}}</code>
- <dd><p>Returns <var>document</var>'s <a>fullscreen element</a>.
-
  <dt><code><var>promise</var> = <var>document</var> . {{Document/exitFullscreen()}}</code>
  <dd><p>Stops <var>document</var>'s <a>fullscreen element</a> from being displayed fullscreen and
  fulfills <var>promise</var> when done.
+
+ <dt><code><var>document</var> . {{DocumentOrShadowRoot/fullscreenElement}}</code>
+ <dd><p>Returns the result of retargeting <var>document</var>'s <a>fullscreen element</a> against
+ <var>document</var>.
```
annevk: I'm not sure about the use of "retargeting" since we never explain it outside of an algorithm. Perhaps we should just keep this vague and say "document's fullscreen element" and "shadow root's fullscreen element" below?

fullscreen.bs
```diff
@@ -283,14 +291,17 @@ these steps:
...
-
 <p>The <dfn attribute for=Document><code>fullscreen</code></dfn> attribute's getter must return
 false if <a>context object</a>'s <a>fullscreen element</a> is null, and true otherwise.
 
 <p class=note>Use
-<a attribute for=Document lt=fullscreenElement><code>document.fullscreenElement</code></a> instead.
+<a attribute for=DocumentOrShadowRoot lt=fullscreenElement><code>document.fullscreenElement</code></a>
+instead.
+
+<p>The <dfn attribute for=DocumentOrShadowRoot><code>fullscreenElement</code></dfn> attribute's
```
annevk: Maybe we should add id=dom-document-fullscreenelement to `<p>`. What do you think @foolip?

fullscreen.bs
```diff
@@ -283,14 +291,17 @@ these steps:
...
 
-<p>The <dfn attribute for=Document><code>fullscreenElement</code></dfn> attribute's getter must
-return <a>context object</a>'s <a>fullscreen element</a>.
-
 <p>The <dfn attribute for=Document><code>fullscreen</code></dfn> attribute's getter must return
 false if <a>context object</a>'s <a>fullscreen element</a> is null, and true otherwise.
 
 <p class=note>Use
-<a attribute for=Document lt=fullscreenElement><code>document.fullscreenElement</code></a> instead.
+<a attribute for=DocumentOrShadowRoot lt=fullscreenElement><code>document.fullscreenElement</code></a>
```
annevk: Seems fine to change that at the same time.

fullscreen.bs
```diff
@@ -283,14 +291,17 @@ these steps:
...
+instead.
+
+<p>The <dfn attribute for=DocumentOrShadowRoot><code>fullscreenElement</code></dfn> attribute's
+getter must return the result of <a>retargeting</a> <a>context object</a>'s
+<a>shadow-including root</a>'s <a>fullscreen element</a> against the <a>context object</a> if the
```
annevk: Yeah that would be a problem. Using "node document" always feels a bit hacky, but it might work here.

fullscreen.bs
foolip: Yeah, that's a good place to put it.

fullscreen.bs
TakayoshiKochi: Done.

fullscreen.bs
TakayoshiKochi: Yeah, as this is in a note, if that looks good to you, it can just say "return xxx's fullscreen element".
Done.

fullscreen.bs
TakayoshiKochi: Done.

fullscreen.bs
TakayoshiKochi: As shadow root doesn't have "node document" (its host has, though) so I changed this sentence into steps.
Could you take another look?

fullscreen.bs
TakayoshiKochi: Yes, the reason for matching `:fullscreen` against shadow host is https://github.com/w3c/webcomponents/issues/180#issuecomment-224233416.

fullscreen.bs
annevk: Why doesn't shadow root have a node document? All nodes have one.

fullscreen.bs
```diff
@@ -283,14 +289,21 @@ these steps:
...
 false if <a>context object</a>'s <a>fullscreen element</a> is null, and true otherwise.
 
-<p class=note>Use
-<a attribute for=Document lt=fullscreenElement><code>document.fullscreenElement</code></a> instead.
+<p class=note>Use {{DocumentOrShadowRoot/fullscreenElement}} instead.
+
+<p>The
+<dfn attribute for=DocumentOrShadowRoot id=dom-document-fullscreenelement><code>fullscreenElement</code></dfn>
+attribute's getter must run the following steps and return the result accordingly.
+<ol>
```
annevk: Newline before `<ol>` and use the phrasing "getter must run these steps:"

fullscreen.bs
```diff
@@ -283,14 +289,21 @@ these steps:
...
+  <li>If the <a>context object</a> is a <a>shadow root</a> and its <a for=DocumentFragment>host</a>
```
annevk: Use `<li><p>` throughout

fullscreen.bs
```diff
@@ -283,14 +289,21 @@ these steps:
...
+  is not <a>connected</a>, return null.</li>
```
annevk: then return*

fullscreen.bs
```diff
@@ -283,14 +289,21 @@ these steps:
...
+  <li>Run <a>retargeting</a> <a>fullscreen element</a> against the <a>context object</a>.
```
annevk: You should store the result in a variable. "Let _candidate_ be the result of ..."

fullscreen.bs
```diff
@@ -505,9 +518,15 @@ properties apply to this pseudo-element either.
 <p>The <dfn id=css-pc-fullscreen><code>:fullscreen</code></dfn> pseudo-class must match any
 <a>element</a> that has its <a>fullscreen flag</a> set.
 
-<p class="note no-backref">This makes it different from the {{Document/fullscreenElement}} API,
-which returns a <a>document</a>'s <a>fullscreen element</a>.
+<p class="note no-backref">This makes it different from the fullscreenElement API, which returns
+returns the topmost fullscreen element..
```
annevk: You droppped the reference to fullscreenElement here. And the double dot looks wrong.

fullscreen.bs
```diff
@@ -505,9 +518,15 @@ properties apply to this pseudo-element either.
...
+
+For each <a>node tree</a>, if the result of <a>retargeting</a> its <a>shadow-including root</a>'s
+<a>fullscreen element</a> against the <a for="tree">root</a> of the <a>node tree</a> is not null and
+the result is in the same <a>tree</a>, <a>:fullscreen</a> pseudo-class must match the result.
 
+<p class="note no-backref">In other words, <a>:fullscreen</a> matches all the non-null results of
+{{DocumentOrShadowRoot/fullscreenElement}} API for each <a>node tree</a>.
```
annevk: It's a little weird to have `:fullscreen` defined in two separate sentences with two separate notes. Also, I'd rather start with the requirement of what `:fullscreen` must match and then have an enumeration (if there are to be multiple options).

fullscreen.bs
TakayoshiKochi: Oops, I misunderstood that DocumentFragment does not have ownerDocument...

fullscreen.bs
TakayoshiKochi: Updated the definition for `:fullscreen` so that the second note be unnecessary.

fullscreen.bs
```diff
@@ -283,14 +289,24 @@ these steps:
...
+attribute's getter must run these steps:
+
+<ol>
+  <li><p>If the <a>context object</a> is a <a>shadow root</a> and its <a for=DocumentFragment>host</a>
+  is not <a>connected</a>, then return null.</li>
+  <li><p>Let <var>candidate</var> be the result of <a>retargeting</a> <a>fullscreen element</a>
+  against the <a>context object</a>.
+  <li><p>If <var>candidate</var> and the <a>context object</a> are in the same tree,
+  return <var>candidate</var>.
+  <li><p>Return null.
```
annevk: Just use a single space for indentation. Newline between the `<li>`s.

fullscreen.bs
```diff
@@ -503,11 +519,12 @@ properties apply to this pseudo-element either.
 <h3 id=:fullscreen-pseudo-class><code>:fullscreen</code> pseudo-class</h3>
 
 <p>The <dfn id=css-pc-fullscreen><code>:fullscreen</code></dfn> pseudo-class must match any
-<a>element</a> that has its <a>fullscreen flag</a> set.
-
-<p class="note no-backref">This makes it different from the {{Document/fullscreenElement}} API,
-which returns a <a>document</a>'s <a>fullscreen element</a>.
+<a>element</a> that has its <a>fullscreen flag</a> set, and the non-null results of
+<a>retargeting</a> <a>fullscreen element</a> against each <a>connected</a> <a for="tree">root</a> of
+<a>node trees</a>.
```
annevk: This still reads weird. It needs to have both the flag set and be the result of retargeting?

Which fullscreen element are we talking about here? What does it belong to?

I'm also not sure that "root of node trees" is a thing or what that actually means.

fullscreen.bs
```diff
@@ -503,11 +519,12 @@ properties apply to this pseudo-element either.
...
 
+<p class="note no-backref">This makes it different from the fullscreenElement API, which returns
```
annevk: You should use `{{DocumentOrShadowRoot/fullscreenElement}}` here.

annevk: Using "the" and "API" makes it more of a sentence, basically. You could also say "the `fullscreenElement` attribute" or some such, but just `fullscreenElement` on its own reads a little weird.

fullscreen.bs
```diff
@@ -502,12 +520,19 @@ properties apply to this pseudo-element either.
 
 <h3 id=:fullscreen-pseudo-class><code>:fullscreen</code> pseudo-class</h3>
 
-<p>The <dfn id=css-pc-fullscreen><code>:fullscreen</code></dfn> pseudo-class must match any
-<a>element</a> that has its <a>fullscreen flag</a> set.
+<p>The <dfn id=css-pc-fullscreen><code>:fullscreen</code></dfn> pseudo-class must match an
+<a>element</a> <var>element</var> if any of the following conditions is true:
```
annevk: "must match any element _element_ for which one of the following is true:" reads a little better I think.

fullscreen.bs
```diff
@@ -502,12 +520,19 @@ properties apply to this pseudo-element either.
...
 
-<p class="note no-backref">This makes it different from the {{Document/fullscreenElement}} API,
-which returns a <a>document</a>'s <a>fullscreen element</a>.
+<ul>
+ <li><p><var>element</var>'s <a>fullscreen flag</a> is set.
+
+ <li><p><var>element</var> is a <a>shadow host</a> and the result of <a>retargeting</a> its
+ <a>node document</a>'s <a>fullscreen element</a> against <var>element</var> is equal to
+ <var>element</var>.
```
annevk: s/is equal to/is/

annevk:
Looks good modulo nits.

fullscreen.bs
```diff
@@ -283,14 +288,27 @@ these steps:
...
+attribute's getter must run these steps:
+
+<ol>
+ <li><p>If the <a>context object</a> is a <a for=/>shadow root</a> and its
+ <a for=DocumentFragment>host</a> is not <a>connected</a>, then return null.</li>
+
+ <li><p>Let <var>candidate</var> be the result of <a>retargeting</a> <a>fullscreen element</a>
+ against the <a>context object</a>.
+
+ <li><p>If <var>candidate</var> and the <a>context object</a> are in the same tree, return
```
foolip: Maybe `<a>tree</a>` here?

fullscreen.bs
foolip: OK, that argument makes sense. I think extra bookkeeping flags will be needed to implement this efficiently, but let's see about that later.

foolip:
LGTM too.

fullscreen.bs
TakayoshiKochi: Done.

fullscreen.bs
TakayoshiKochi: Done.

fullscreen.bs
TakayoshiKochi: Done.

fullscreen.bs
```diff
@@ -283,14 +288,27 @@ these steps:
...
 
-<p>The <dfn attribute for=Document><code>fullscreenElement</code></dfn> attribute's getter must
-return <a>context object</a>'s <a>fullscreen element</a>.
-
 <p>The <dfn attribute for=Document><code>fullscreen</code></dfn> attribute's getter must return
 false if <a>context object</a>'s <a>fullscreen element</a> is null, and true otherwise.
 
-<p class=note>Use
-<a attribute for=Document lt=fullscreenElement><code>document.fullscreenElement</code></a> instead.
+<p class=note>Use {{DocumentOrShadowRoot/fullscreenElement}} instead.
```
annevk: the `fullscreenElement` attribute* (this is why it was document.fullScreenElement before I think)

fullscreen.bs
```diff
@@ -283,14 +288,27 @@ these steps:
...
+attribute's getter must run these steps:
+
+<ol>
+ <li><p>If the <a>context object</a> is a <a for=/>shadow root</a> and its
+ <a for=DocumentFragment>host</a> is not <a>connected</a>, then return null.</li>
+
+ <li><p>Let <var>candidate</var> be the result of <a>retargeting</a> <a>fullscreen element</a>
+ against the <a>context object</a>.
+
+ <li><p>If <var>candidate</var> and the <a>context object</a> are in the same <a>tree</a>, return
```
annevk: then return

fullscreen.bs
TakayoshiKochi: Done.

fullscreen.bs
TakayoshiKochi: Done.

TakayoshiKochi:
Thanks for the review!

annevk:
Thanks for the great work on the patch!

PR #55, opened 2016-09-16
To fullscreen/unfullscreen, do not take document as an argument

TakayoshiKochi:
The definitions used to use document as argument to fullscreen or
unfullscreen an element, but document could be induced from the
element's node document.


TakayoshiKochi:
@annevk @foolip PTAL

annevk:
Thanks!

TakayoshiKochi:
Thanks!

(next time I'll include html as well)

PR #57, opened 2016-09-27
Move the element type check out of the ready check

foolip:
There is no need to check this again in the animation frame task.

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -184,12 +184,6 @@ if all of the following are true, and false otherwise:
 
 <ul>
 
- <li><p><var>element</var>'s <a for=Element>namespace</a> is the <a>HTML namespace</a> or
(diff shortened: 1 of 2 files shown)


fullscreen.bs
```diff
@@ -208,14 +202,20 @@ these steps:
  <li><p>Let <var>promise</var> be a new promise.
 
  <li>
-  <p>If any of the following conditions are true, set <var>error</var> to true:
+  <p>If any of the following conditions are false, set <var>error</var> to true:
```
annevk: then set*

annevk:
LGTM with that.

foolip:
I also fixed two other instances, can you confirm that is what you had in mind?

foolip:
Test: https://github.com/w3c/web-platform-tests/pull/3830

PR #60, opened 2016-10-24
Deny request on invisible document

### LINKED ISSUE #58: Fullscreen request on invisible document should be denied
I suppose that exposes security risks that attacker can open a background window and put it into fullscreen without having user notice it. And then when user switches window, it may start spoofing.

upsuper:
Fixes #58.

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -20,6 +20,7 @@ Boilerplate: omit feedback-header, omit conformance, omit index, omit idl-index
 <pre class=link-defaults>
 spec:dom; type:dfn; for:/; text:element
 spec:dom; type:interface; text:Document
+spec:page-visibility; type:dfn; text:hidden
 </pre>
 
 <pre class=anchors>
@@ -194,6 +195,8 @@ if all of the following are true, and false otherwise:
  <li><p><var>element</var>'s <a>node document</a> is <a>allowed to use</a> the feature indicated by
  attribute name <code>allowfullscreen</code>.
  <!-- cross-process, recursive -->
+
+ <li><p><var>element</var>'s <a>node document</a>'s <a>hidden</a> attribute returns false.
 </ul>
 
 <p>The <dfn method for=Element><code>requestFullscreen()</code></dfn> method, when invoked, must run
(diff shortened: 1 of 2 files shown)


annevk:
I think we should instead invoke https://w3c.github.io/page-visibility/#dfn-steps-to-determine-the-visibility-state directly.

IDL attributes is generally what is exposed to JavaScript, whereas specification algorithms should operate on what is exposed to IDL. You can contrast this with C++, where we'd also generally use some underlying concept.

annevk:
Also, I'd like @foolip to sign off on the general idea.

foolip:
Can you confirm that the `document.hidden` attribute actually true false for background windows? I assumed that would require involving the window manager some how, and that as actually implemented it only returns true for background tabs.

upsuper:
It returns `false` for background window, but `true` for background tab. It seems in Chrome, background tab wouldn't be allowed to enter fullscreen either, how does that work?

We stop background window from entering fullscreen via detecting whether the current document has focus, I think. Chrome seems to allow background window to enter fullscreen. Do we want to add the focus check to the element ready check instead?

foolip:
Is the example in https://github.com/whatwg/fullscreen/issues/58#issuecomment-250416978 the only thing that this is supposed to prevent? In my testing, the newly opened window is not hidden, which I think matches your testing.

Presumably it's easy to also give the new window focus, but is the goal to prevent all new windows from going fullscreen, or just some cases?

upsuper:
I think the goal is to prevent a background window from going fullscreen silently, so probably we want to check whether the document or the top browsing context of document currently has focus.

foolip:
Can you test if `document.activeElement` becomes null in non-active windows, and if there are any events fired? Given http://stackoverflow.com/questions/1060008/is-there-a-way-to-detect-if-a-browser-window-is-not-currently-active I wouldn't be surprised if there's no web exposed API to tell the difference.

foolip:
Ping @upsuper

upsuper:
`document.activeElement` doesn't become null. `focus` and `blur` would be fired when switching between windows, but changing the focus to address bar also triggers `blur`.

If it is not allowed to trigger fullscreen asynchronously, the top level document should always get focused before `requestFullscreen` is called, unless a document from another window requests that.

So should we allow a foreground document to put a background document into fullscreen? Given that window cannot raise itself in general, I don't see how allowing that could be useful.

foolip:
> So should we allow a foreground document to put a background document into fullscreen? Given that window cannot raise itself in general, I don't see how allowing that could be useful.

I agree that doesn't sound useful. What should the spec say to prevent that? One thing that might work is a document-bound user gesture concept, so that it's only possible to request fullscreen for the document that was clicked. Is that what you had in mind?

PR #62, opened 2016-11-18
Editorial: use the Infra Standard

annevk:
Also add a Makefile for convenience and update the README.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,5 @@
+local: fullscreen.bs
+	bikeshed
+
+remote: fullscreen.bs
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -0,0 +1,5 @@
+local: fullscreen.bs
+	bikeshed
+
+remote: fullscreen.bs
+	curl https://api.csswg.org/bikeshed/ -f -F file=@fullscreen.bs > fullscreen.html
```
foolip: Yay, I've often had to copy-paste this :)

annevk: Note that this version doesn't give you warnings and such. But you do get to check the output.

foolip: Good enough for trivial changes :)

PR #64, opened 2016-11-22
Decide whether to resize synchronously in requestFullscreen()

### LINKED ISSUE #33: What should happen if an element is moved to another document right after requestFullscreen
Unless I'm missing something, this should just work per spec, but feels odd.
For other cases there is "Whenever the removing steps run with an oldNode, run these steps"

### LINKED ISSUE #63: Decide on wanted behavior for request+exit or exit+request together
Working on Fullscreen in Blink, the situation has come up where we're transition in or out of fullscreen when a call to go in the opposite direction happens. Test case:

```HTML
<!doctype html>
<div>
 <button>request</button>
 <button>exit</button>
 <button>request+exit</button>
 <button>exit+request</button>
</div>
</div>
<script>
var div = document.querySelector('div');
div.addEventListener('click', function(event) {
  if (event.target.localName != 'button')
    return;
  for (var action of event.target.textContent.split('+')) {
    if (action == 'request') {
      if (div.requestFullscreen) {
        div.requestFullscreen();
      } else if (div.webkitRequestFullscreen) {
        div.webkitRequestFullscreen();
      } else {
        div.mozRequestFullScreen();
      }
    } else if (action == 'exit') {
      if (document.exitFullscreen) {
        document.exitFullscreen();
      } else if (document.webkitExitFullscreen) {
        document.webkitExitFullscreen();
      } else {
        document.mozCancelFullScreen();
      }
    } else {
      throw 'unknown action';
    }
  }
});
</script>
```

Current behavior:
<table>
<tr>
 <th></th>
 <th>request+exit</th>
 <th>exit+request</th>
 <th>request then request+exit</th>
 <th>request then exit+request</th>
<tr>
 <th>Spec'd</th>
 <th>enter</th>
 <th>enter</th>
 <th>???</th>
 <th>???</th>
<tr>
 <th>Chrome 55</th>
 <th>enters+exits quickly</th>
 <th>enters</th>
 <th>exits</th>
 <th>exits</th>
<tr>
 <th>Edge 14</th>
 <th>enters+exits quickly</th>
 <th>enters</th>
 <th>exits</th>
 <th>exits+enters quickly</th>
<tr>
 <th>Firefox 51</th>
 <th>enters</th>
 <th>enters</th>
 <th>exits</th>
 <th>exits</th>
</table>

Safari 10 behaved strangely when testing in BrowserStack, seemingly hanging, so I don't trust what I saw.

[Exit fullscreen](https://fullscreen.spec.whatwg.org/#exit-fullscreen) has an early return in the spec, which makes the first two cases simple. For the second two, it's not obvious well defined, because two "resizes" are racing "in parallel."

@upsuper @jernoble

foolip:
Fixes #63. It does not fix #33, although to fix it would now involve
saying "If both error and resize are true, then ..." in the animation
frame task.

***
[Preview](https://fullscreen.spec.whatwg.org/branch-snapshots/requestFullscreen-resize/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fullscreen/cbdc4f7...7ab6a5d.html)

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -213,18 +213,26 @@ these steps:
    element. [[!SVG]] [[!MATHML]]
 
    <li><p>The <a>fullscreen element ready check</a> for <var>pending</var> returns true.
+   <!-- cross-process; check is only needed on pending as it is recursive already -->
 
    <li><p><a>Fullscreen is supported</a>.
 
    <li><p>This algorithm is <a>allowed to request fullscreen</a>.
   </ul>
 
+ <li><p>Let <var>topLevelDoc</var> be <var>pendingDoc</var>'s <a>top-level browsing context</a>'s
+ <a>active document</a>.
+ <!-- cross-process -->
+
+ <li><p>Let <var>resize</var> be true, if <var>error</var> is false and <var>topLevelDoc</var>'s
+ <a>fullscreen element</a> is null, and false otherwise.
+ <!-- cross-process -->
+
  <li><p>Return <var>promise</var>, and run the remaining steps <a>in parallel</a>.
 
- <li><p>If <var>error</var> is false: Resize <var>pendingDoc</var>'s
- <a>top-level browsing context</a>'s <a>active document</a>'s viewport's dimensions to match the
- dimensions of the screen of the output device. Optionally display a message how the end user can
- revert this.
+ <li><p>If <var>resize</var> is true: Resize <var>topLevelDoc</var>'s viewport's dimensions to match
+ the dimensions of the screen of the output device. Optionally display a message how the end user
+ can revert this.
  <!-- cross-process -->
 
  <li>
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -247,11 +247,28 @@ these steps:
 
      <li><p>The <a>fullscreen element ready check</a> for <var>pending</var> returns true.
      <!-- cross-process; check is only needed on pending as it is recursive already -->
+
+     <li><p><var>topLevelDoc</var>'s viewport matches the dimensions of the screen.
+     <!-- cross-process -->
     </ul>
 
-   <li><p>If <var>error</var> is true, <a>fire an event</a> named <code>fullscreenerror</code> on
-   <var>pendingDoc</var>, reject <var>promise</var> with a <code>TypeError</code> exception, and
-   terminate these steps.
+    <p class=note>For example, <var>pending</var> might have been moved to another document,
+    <code>allowfullscreen</code> attributes might have been removed, or <var>topLevelDoc</var> might
+    have exited fullscreen since {{Element/requestFullscreen()}} was invoked.
+
+   <li>
+    <p>If <var>error</var> is true:
+
+    <ol>
+     <li><p><a>Fire an event</a> named <code>fullscreenerror</code> on <var>pendingDoc</var>.
+
+     <li><p>Reject <var>promise</var> with a <code>TypeError</code> exception.
+
+     <li><p>Terminate these steps and run the following step <a>in parallel</a>.
+
+     <li><p>If <var>resize</var> is true, resize <var>topLevelDoc</var>'s viewport to its "normal"
+     dimensions.
+    </ol>
 
    <li><p>Let <var>fullscreenElements</var> be an <a>ordered set</a> initially consisting of
    <var>pending</var>.


fullscreen.bs
```diff
@@ -217,11 +225,26 @@ these steps:
...
+      <a>active document</a> is <var>topLevelDoc</var>.
+      <!-- cross-process -->
+
+      <p class=note><var>pending</var> may have moved to another <a>top-level browsing context</a>
+      since the request.
+    </ul>
+
+   <li><p>If <var>error</var> is true, <a>fire an event</a> named <code>fullscreenerror</code> on
    <var>pending</var>'s <a>node document</a>, reject <var>promise</var> with a
    <code>TypeError</code> exception, and terminate these steps.
```
upsuper: This actually reminds me another issue with the old spec text: if the fullscreen request triggers resizing, but fails here, the viewport would keep fullscreen, while the content would not be in fullscreen state, which seems to be unfortunate.

Rather than simply "terminate these steps", it should probably also say "If *resize* is true, resize *topLevelDoc*’s viewport to its 'normal' dimensions" before.

(This actually caused a vulnerability in Gecko which is fixed recently. And I didn't realize this is a spec issue.)

That would still be tricky, though. If one triggers two fullscreen requests continuously, and the latter one fails the check here, then boom. But I guess it is better than leaving window in fullscreen while content is not.

fullscreen.bs
```diff
@@ -217,11 +225,26 @@ these steps:
   <p>As part of the next <a>animation frame task</a>, run these substeps:
 
   <ol>
-   <li><p>If either <var>error</var> is true or the <a>fullscreen element ready check</a> for
-   <var>pending</var> returns false, <a>fire an event</a> named <code>fullscreenerror</code> on
+
+   <li>
+    <p>If any of the following conditions are false, then set <var>error</var> to true:
```
upsuper: I might be wrong, but I think "any" should use "is" rather than "are" here?

foolip: I don't think so, because conditions is plural. "If any following condition is false" would be grammatically correct, but less clear.

fullscreen.bs
foolip: Yes, both in the case where the element was removed after the request and if it was moved to another window as in https://github.com/whatwg/fullscreen/issues/33, we might have gone fullscreen but there's nothing to make the fullscreen element. Calling "fully exit fullscreen" would do nothing, so I think your suggestion is the best here. I'll try an edit.

foolip:
@annevk, could you also review?

fullscreen.bs
```diff
@@ -199,29 +199,63 @@ these steps:
...
+
+   <li>
+    <p>If <var>error</var> is true, run these subsubsteps:
+
+    <ol>
+     <li><p><a>Fire an event</a> named <code>fullscreenerror</code> on <var>pendingDoc</var>.
+
+     <li><p>Reject <var>promise</var> with a <code>TypeError</code> exception.
+
+     <li><p>Terminate these steps and run the remaining subsubsteps <a>in parallel</a>.
```
foolip: @annevk, help with better wording?

fullscreen.bs
```diff
@@ -199,29 +199,63 @@ these steps:
...
+
+     <li><p>If <var>resize</var> is true, resize <var>topLevelDoc</var>'s viewport to its "normal"
```
foolip: This is still not quite right. There could be a fullscreen element, in which case one would also have to fiddle with the top layer. Ideas on the minimal amount of fiddling required to end up in a good state appreciated.

fullscreen.bs
annevk: Maybe just "If resize is true, in parallel, resize ..."?

annevk: Oh, the parallel step is not quite right...

fullscreen.bs
```diff
@@ -199,29 +199,63 @@ these steps:
...
 
    <li><p>This algorithm is <a>allowed to request fullscreen</a>.
   </ul>
 
+ <li><p>Let <var>pendingDoc</var> be <var>pending</var>'s <a>node document</a>.
+
+ <li><p>Let <var>topLevelDoc</var> be <var>pendingDoc</var>'s <a>top-level browsing context</a>'s
+ <a>active document</a>. <!-- cross-process -->
+
+ <li><p>Let <var>resize</var> be true if <var>error</var> is false and <var>topLevelDoc</var>'s
```
annevk: Comma after true.

fullscreen.bs
```diff
@@ -199,29 +199,63 @@ these steps:
...
+ <a>fullscreen element</a> is null, and false otherwise.
+ <!-- cross-process -->
+
  <li><p>Return <var>promise</var>, and run the remaining steps <a>in parallel</a>.
 
- <li><p>If <var>error</var> is false: Resize <var>pending</var>'s
- <a>top-level browsing context</a>'s <a>active document</a>'s viewport's dimensions to match the
- dimensions of the screen of the output device. Optionally display a message how the end user can
- revert this.
+ <li><p>If <var>resize</var> is true, resize <var>topLevelDoc</var>'s viewport to match the
```
annevk: then resize*

fullscreen.bs
foolip: The trick is how to both stop the steps in the animation frame task and to capture the fact that the resize doesn't happen synchronously within the animation frame task. But let's figure out the right behavior first.

fullscreen.bs
foolip: It would be a case like this:
```JavaScript
button.addEventListener('click', () => {
  a.requestFullscreen();
  b.requestFullscreen()
  b.remove();
});
```

Both requests will try to resize, and per spec it's a racy situation, but let's assume that after the resize the animation frame task for *a* runs first and *a* becomes the fullscreen element. Then *b* will end up in this situation, exiting fullscreen, but *a* will still be the fullscreen element.

foolip:
Given https://github.com/whatwg/fullscreen/pull/64#discussion_r89648822 and that `exitFullscreen()` has a similar race, I think that the approach in the PR won't quite work.

In order to make it well defined what happens with racing resizes and which order all the promises are resolved, I think we'll need to have some state on the top-level browsing context, or *some* kind of queue of requests. For at least exits, I think that all requests will also have be handled in a single animation frame task following the resize, otherwise some promises can't be resolved in the same tasks where the corresponding events are fired.

What do you think, can it be simpler and still completely well defined?

upsuper:
Gecko has a global queue for pending fullscreen requests when the fullscreen state cannot be applied immediately. (It is especially important for us because we have a transition animation.)

And our algorithm is roughly:
* If *resize* is true or there is any fullscreen request in the queue has the same top level document, push the fullscreen request into the queue, and trigger the resizing.
  * When the viewport gets resized, iterates the queue, and for each fullscreen request whose document's top level document's viewport is what we just resized:
    * remove the request from the queue, and
    * apply the fullscreen request.
  * If no fullscreen request is actually applied (which indicates the top level document's fullscreen element is still null), resize the viewport back to its normal dimension.
* Otherwise, apply the fullscreen request.

Having written this down, I wonder whether this is the only reason we reject when element is moved to a different document. If so, I guess we probably don't really need that restriction.

We should probably be recording the top level document instead, otherwise it would still be an issue that the requesting document can be moved to a different top level document.

foolip:
Thanks, @upsuper, that's very helpful. Is this queue in the same process/thread as the rendering engine, i.e. can you inspect its state synchronously in `requestFullscreen()`?

Are you currently synchronizing "apply the fullscreen request" with animation frames, or does it happen whenever the resize is reported to be done?

> Having written this down, I wonder whether this is the only reason we reject when element is moved to a different document. If so, I guess we probably don't really need that restriction.

Yes, keeping track only of *topLevelDoc* would for for #33. Using *pendingDoc* amounts to a pretty harmless restriction though, so I wouldn't mind aligning with that if it'd be simpler for you.

> We should probably be recording the top level document instead, otherwise it would still be an issue that the requesting document can be moved to a different top level document.

I don't think it's possible to move documents, here's a test that shows that moving an iframe causes its document to change: https://software.hixie.ch/utilities/js/live-dom-viewer/saved/4693

In addition to all of the above, can you elaborate on how exiting works? Do exits also end up in the same queue, and how are competing requests and exists handled? I would like the spec to match Gecko in https://github.com/whatwg/fullscreen/issues/63, so if you have a good model we can just copy it exactly :)

upsuper:
> Is this queue in the same process/thread as the rendering engine, i.e. can you inspect its state synchronously in `requestFullscreen()`?

Yes, yes.

> Are you currently synchronizing "apply the fullscreen request" with animation frames, or does it happen whenever the resize is reported to be done?

Not really. In Gecko, in case `resize` is true, "apply the fullscreen request" happens as soon as we get the message from the window manager that the window size has been changed, but the fullscreen event and resize event is delayed to the next animation frame. If the `resize` is false, we queue a task to "apply the fullscreen request", and dispatch the event in the next animation frame.

I think the difference between this behavior and doing everything in the next animation frame is probably observable, but should be hard to detect in practice, so it may not worth fixing in either spec side or Gecko's impl.

FWIW, I guess we would fix this gap in Gecko when we implement `Promise` for `requestFullscreen`.

> Yes, keeping track only of topLevelDoc would for for #33. Using pendingDoc amounts to a pretty harmless restriction though, so I wouldn't mind aligning with that if it'd be simpler for you.
> I don't think it's possible to move documents, here's a test that shows that moving an iframe causes its document to change: https://software.hixie.ch/utilities/js/live-dom-viewer/saved/4693

Given these, I would be happy that the spec just uses the pending document :)

> In addition to all of the above, can you elaborate on how exiting works? Do exits also end up in the same queue, and how are competing requests and exists handled? I would like the spec to match Gecko in #63, so if you have a good model we can just copy it exactly :)

No, exits just go immediately, and they are even synchronous if `resize` is false... (I probably should at least make them asynchronous in the same way `requestFullscreen` does.) I considered establishing a standalone exiting queue for resolving `Promise`s, but haven't got a clear idea about the details.

Oh, and in `requestFullscreen`, "set resize" happens asynchronously as well... which is probably bad. Given this inconsistency on behavior between request and exit fullscreen, we probably don't want to spec the exactly same algorithm.

I don't think we have any explicit handling of competing requests and exits. I think the two phases resize ("set `resize`" synchronously and "resize viewport" asynchronously) should be fine enough to spec.

foolip:
I've rebased this now as it came up in https://github.com/whatwg/fullscreen/issues/88, and split it into two commits. This is taking the third option from https://github.com/whatwg/fullscreen/issues/63#issuecomment-261948581 but I haven't fully convinced myself this is the best path, since it makes the cross-process considerations harder.

I'm still working on https://github.com/whatwg/fullscreen/pull/86 and there are tricky things about "exit fullscreen" to sort out there, so I'm going to finish that before coming back to this.

foolip:
Closing this. https://github.com/whatwg/fullscreen/issues/63 still needs fixing, but probably not like this. Not allowing competing requests would be a simpler model.

PR #65, opened 2016-12-14
Give "exit fullscreen" a fully flag and define "fully exit fullscreen" using it

foolip:
The previous definition of fully exit fullscreen could unfullscreen
elements in top layer synchronously. If those elements were iframes,
the contentDocument would get handled in a following animation frame
task, leaving everything in an odd state in the interim.

Exposing the concept of fully exiting to scripts is https://github.com/whatwg/fullscreen/issues/70.

***
[Preview](https://fullscreen.spec.whatwg.org/branch-snapshots/fully-exit-fullscreen/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fullscreen/cbdc4f7...9330439.html)

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -85,18 +85,6 @@ its <a>node document</a>'s <a>top layer</a>.
 
 <hr>
 
-<p>To <dfn>fully exit fullscreen</dfn> a <a for=/>document</a> <var>document</var>, run these steps:
-
-<ol>
- <li><p>If <var>document</var>'s <a>fullscreen element</a> is null, terminate these steps.
-
- <li><p><a lt="unfullscreen an element">Unfullscreen elements</a> whose <a>fullscreen flag</a> is
- set, within <var>document</var>'s <a>top layer</a>, except for <var>document</var>'s
- <a>fullscreen element</a>.
-
- <li><p><a>Exit fullscreen</a> <var>document</var>.
-</ol>
-
 <p>Whenever the <a>removing steps</a> run with a <var>removedNode</var>, run these steps:
 
 <ol>
@@ -109,7 +97,7 @@ its <a>node document</a>'s <a>top layer</a>.
 
   <ol>
    <li><p>If <var>node</var> is its <a>node document</a>'s <a>fullscreen element</a>,
-   <a>exit fullscreen</a> that <a for=/>document</a>.
+   <a>exit fullscreen</a> that <a for=/>document</a> with <var>fullyFlag</var> unset.
 
    <li><p>Otherwise, <a lt="unfullscreen an element">unfullscreen <var>node</var></a> within its
    <a>node document</a>.
@@ -117,7 +105,7 @@ its <a>node document</a>'s <a>top layer</a>.
 </ol>
 
 <p>Whenever the <a>unloading document cleanup steps</a> run with a <var>document</var>,
-<a>fully exit fullscreen</a> <var>document</var>.
+<a>exit fullscreen</a> <var>document</var> with <var>fullyFlag</var> set.
 
 <hr>
 
@@ -351,7 +339,8 @@ could be an open <{dialog}> element.
  in which case that document will still remain in fullscreen.
 </ol>
 
-<p>To <dfn>exit fullscreen</dfn> a <a for=/>document</a> <var>doc</var>, run these steps:
+<p>To <dfn>exit fullscreen</dfn> a <a for=/>document</a> <var>doc</var> with
+<var>fullyFlag</var>, run these steps:
 
 <ol>
  <li><p>Let <var>promise</var> be a new promise.
@@ -393,8 +382,8 @@ could be an open <{dialog}> element.
 
    <li><p>If <var>resize</var> is true and <var>topLevelDoc</var> is either not in
    <var>exitDocs</var>, or not a <a>simple fullscreen document</a>,
-   <a>fully exit fullscreen</a> <var>topLevelDoc</var>, reject <var>promise</var> with a
-   <code>TypeError</code> exception, and terminate these steps.
+   <a>exit fullscreen</a> <var>topLevelDoc</var> with <var>fullyFlag</var> set, reject
+   <var>promise</var> with a <code>TypeError</code> exception, and terminate these steps.
    <!-- cross-process -->
 
    <li><p>Let <var>descendantDocs</var> be an <a>ordered set</a> consisting of <var>doc</var>'s
@@ -406,8 +395,16 @@ could be an open <{dialog}> element.
    <a lt="unfullscreen a document">unfullscreen <var>descendantDoc</var></a>.
    <!-- cross-process -->
 
-   <li><p><a>For each</a> <var>exitDoc</var> in <var>exitDocs</var>,
-   <a lt="unfullscreen an element">unfullscreen</a> <var>exitDoc</var>'s <a>fullscreen element</a>.
+   <li>
+    <p><a>For each</a> <var>exitDoc</var> in <var>exitDocs</var>:
+
+    <ol>
+     <li><p>If <var>exitDoc</var> is <var>doc</var> and <var>fullyFlag</var> is set, then
+     <a lt="unfullscreen a document">unfullscreen <var>doc</var></a>.
+
+     <li><p>Otherwise, <a lt="unfullscreen an element">unfullscreen</a> <var>exitDoc</var>'s
+     <a>fullscreen element</a>.
+    </ol>
 
    <li><p><a>For each</a> <var>descendantDoc</var> in <var>descendantDocs</var>,
    <a>fire an event</a> named <code>fullscreenchange</code> on <var>descendantDoc</var>.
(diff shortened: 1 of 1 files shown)


foolip:
This is a big rough, but the structure matches exactly what I found myself having to do in https://codereview.chromium.org/2573773002/ in order to make sense of fully exit fullscreen with the timing changes to modify the fullscreen element stack (soon top layer) at animation frame timing.

I have doubts about whether to expose this in the API or not. It would be great for testing and I even have the tests, but is it also useful for web authors?

annevk:
I guess it might be useful if you "own" the entire application stack. But what would this do inside an `<iframe>`? Would it actually fully exit?

foolip:
You changed https://fullscreen.spec.whatwg.org/#fully-exit-fullscreen to only fully exit in the document for which it's called in 3243119d027a8ff5b80998eb1f17f8eba148a346. I assumed that was intentional? Do you think people would expect `iframe.contentDocument.exitFullscreen({ fully: true })` to exit all the way up?

annevk:
It does seem like that was intentional. I guess as long as you're same-origin you can grab the same document as UX would when fully exiting.

fullscreen.bs
```diff
@@ -365,7 +362,9 @@ attribute's getter must run these steps:
    <a lt="unfullscreen a document">unfullscreen <var>descendantDoc</var></a>.
    <!-- cross-process -->
 
-   <li><p>For each <var>exitDoc</var> in <var>exitDocs</var>, in order,
+   <li><p>For each <var>exitDoc</var> in <var>exitDocs</var>, in order: If <var>exitDoc</var> is
```
foolip: Suggestions also welcome here. Would it be better have exitDocs contain only the documents that should fully exit, and then handle the document above separate? (That is currently "If exitDocs’s last document has a browsing context container, append that browsing context container’s node document to exitDocs.")

fullscreen.bs
```diff
@@ -309,7 +305,8 @@ attribute's getter must run these steps:
  <li><p>Return <var>docs</var>.
 </ol>
 
-<p>To <dfn>exit fullscreen</dfn> a <a for=/>document</a> <var>doc</var>, run these steps:
+<p>To <dfn>exit fullscreen</dfn> a <a for=/>document</a> <var>doc</var> with an optional
+<var>fully flag</var>, run these steps:
```
annevk: Just remove "optional"?

fullscreen.bs
annevk: Not sure, whatever matches the code. But I would use a sublist after "in order" to deal with the if/otherwise situation in two list items rather than jamming it all together.

upsuper:
> Exposing the concept of fully exiting to scripts makes it easier to test, although this is not central to the change itself.

If exposing it is not central of this change, could we split this into a separate issue? It seems to me this could be more controversial than fixing the timing issue.

foolip:
I've now finished the changes and think they are ready for review. There is one existing problem that I'll comment on.

fullscreen.bs
```diff
@@ -340,23 +330,26 @@ attribute's getter must run these steps:
...
    <a>browsing context container</a>, append that <a>browsing context container</a>'s
    <a>node document</a> to <var>exitDocs</var>.
    <!-- cross-process -->
 
+   <li><p>If <var>resize</var> is true, then: while <var>exitDocs</var>'s last <a for=/>document</a>
+   has a <a>browsing context container</a>, append that <a>browsing context container</a>'s
+   <a>node document</a> to <var>exitDocs</var>.
+   <!-- cross-process -->
+
    <li><p>Let <var>descendantDocs</var> be an ordered set consisting of <var>doc</var>'s
```
foolip: Using *doc* here seems like a problem before and after this change. Before, the fullscreen element in one of *exitDocs* may have been another iframe, since we have no hierarchy restrictions now. The wrong part of the frame tree is then going to get its fullscreen state cleared.

With this change, there's the additional problem that if *fully flag* is set, we'll exit fully in all those documents, and there may be a bunch of frames that need state cleared that don't get it.

I'm thinking that maybe the synchronous intro of "exit fullscreen" should instead determine a single *exitDoc*, from which one or all fullscreen top layer elements will be removed depending on the *fully flag*. Then there will be no "collecting documents to unfullscreen" in the animation frame task, only an attempt to make it so. Would that kind of model work? Should I work it into this PR or not?

annevk: @upsuper can probably answer the model question much better than I can.

fullscreen.bs
```diff
@@ -100,15 +88,15 @@ it from its <a>node document</a>'s <a>top layer</a>.
 
   <ol>
    <li><p>If <var>node</var> is its <a>node document</a>'s <a>fullscreen element</a>,
-   <a>exit fullscreen</a> that <a for=/>document</a>.
+   <a>exit fullscreen</a> that <a for=/>document</a> with the <var>fully flag</var> unset.
```
annevk: I would prefer _fullyFlag_ throughout.

fullscreen.bs
```diff
@@ -366,8 +359,16 @@ attribute's getter must run these steps:
    <a lt="unfullscreen a document">unfullscreen <var>descendantDoc</var></a>.
    <!-- cross-process -->
 
-   <li><p>For each <var>exitDoc</var> in <var>exitDocs</var>, in order,
-   <a lt="unfullscreen an element">unfullscreen</a> <var>exitDoc</var>'s <a>fullscreen element</a>.
+   <li>
+    <p>For each <var>exitDoc</var> in <var>exitDocs</var>, in order:
```
annevk: We could make this use https://infra.spec.whatwg.org/#list-iterate and no longer have to say "in order", but could also do that later on.

fullscreen.bs
```diff
@@ -340,23 +330,26 @@ attribute's getter must run these steps:
...
    <li><p>If <var>exitDocs</var> is the empty set, append <var>doc</var> to <var>exitDocs</var>.
 
    <li><p>If <var>exitDocs</var>'s last <a for=/>document</a> has a
    <a>browsing context container</a>, append that <a>browsing context container</a>'s
    <a>node document</a> to <var>exitDocs</var>.
    <!-- cross-process -->
 
+   <li><p>If <var>resize</var> is true, then: while <var>exitDocs</var>'s last <a for=/>document</a>
+   has a <a>browsing context container</a>, append that <a>browsing context container</a>'s
+   <a>node document</a> to <var>exitDocs</var>.
```
upsuper: This would not work, because we only call *unfullscreen an element* on documents in *exitDocs*.

foolip:
Sorry for leaving this sitting for so long, I was fighting fullscreen regressions and being distracted. This will probably need some tweaking after we agree on some change (or not) in https://github.com/whatwg/fullscreen/issues/74 so I'll leave this sitting and ping again when it's ready for review again.

fullscreen.bs
foolip: The algorithm was changed in https://github.com/whatwg/fullscreen/pull/72 and in https://github.com/whatwg/fullscreen/pull/81 I've tried to make it clear what the return value of *collect documents to unfullscreen* means.

So, given no hierarchy restrictions I think that the problem remains, but I'm not going to try to tackle it in this PR.

foolip:
I have reworked and rebased this on top of https://github.com/whatwg/fullscreen/pull/81, which gets rid of some of the changes I had. It's not really possible to write tests for "fully exit fullscreen" with web-exposed APIs currently AFAIK, I have two tests in Blink that depend on something funny about window.open that I don't think is per spec.

So, I think testing the algorithm is blocked on https://github.com/whatwg/fullscreen/issues/70 and will file a wpt issue about missing test coverage.

foolip:
@upsuper, can you give this another look? On top of this I have a WIP change to revamp how animation frame tasks are used to solve https://github.com/whatwg/fullscreen/issues/74, which I need to resolve in order to try relanding https://crbug.com/402376.

foolip:
(Only review the second commit, and I'll rebase once #81 is merged.)

fullscreen.bs
```diff
@@ -109,15 +97,15 @@ its <a>node document</a>'s <a>top layer</a>.
...
+   <a>exit fullscreen</a> that <a for=/>document</a> with <var>fullyFlag</var> unset.
 
    <li><p>Otherwise, <a lt="unfullscreen an element">unfullscreen <var>node</var></a> within its
    <a>node document</a>.
   </ol>
 </ol>
 
 <p>Whenever the <a>unloading document cleanup steps</a> run with a <var>document</var>,
-<a>fully exit fullscreen</a> <var>document</var>.
+<a>exit fullscreen</a> <var>document</var> with <var>fullyFlag</var> set.
```
foolip: I've realized this is actually a bit broken. Exit fullscreen is async and recursive, and
* The async bit means that things should happen to the document after the resize, when it's (probably) already not the active document. In particular, events should be fired, that aren't fired in at least Firefox in https://github.com/w3c/web-platform-tests/pull/5887
* https://html.spec.whatwg.org/multipage/browsers.html#unload-a-document itself recurses into descendant browsing contexts, so there would be more exiting going on than necessary.

I think it would actually be best to just unfullcreen the document with no events at all here, and will change the PR to do that.

foolip: Sigh, no, that doesn't work either. I've filed https://github.com/whatwg/fullscreen/issues/82 and ask to declare this out of scope for this refactoring.

foolip:
This issue would actually become moot with https://github.com/whatwg/fullscreen/pull/91, because then it's impossible for an iframe to be anything except the topmost fullscreen element in top layer.

foolip:
Closing, after https://github.com/whatwg/fullscreen/pull/128 and related discussion I don't think we'll do this.

PR #66, opened 2016-12-14
Check the fullscreen element in the animation frame task for exit

foolip:
The change to fulfill instead of reject is a drive-by for consistency.

The problem in the animation frame task was that if there are two
exitFullscreen() calls for the same document in the same animation frame
task, and an ancestor document has more than one fullscreen element in
its top layer, then the first task could pop one of them, and the second
another, potentially exiting out of fullscreen entirely.

With this change, exitFullscreen() is still not idempotent as doc's top
layer may be emptied where just one call would not, but at least it
stops when obviously silly things are happening.

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -314,8 +314,8 @@ attribute's getter must run these steps:
 <ol>
  <li><p>Let <var>promise</var> be a new promise.
 
- <li><p>If <var>doc</var>'s <a>fullscreen element</a> is null, reject <var>promise</var> with a
- <code>TypeError</code> exception, and return <var>promise</var>.
+ <li><p>If <var>doc</var>'s <a>fullscreen element</a> is null, fulfill <var>promise</var> with
+ undefined, and return <var>promise</var>.
 
  <li><p>Let <var>resize</var> be false.
 
@@ -339,6 +339,9 @@ attribute's getter must run these steps:
   <p>As part of the next <a>animation frame task</a>, run these substeps:
 
   <ol>
+
+   <li><p>If <var>doc</var>'s <a>fullscreen element</a> is null, fulfill <var>promise</var> with
+   undefined, and terminate these steps.
 
    <li><p>Let <var>exitDocs</var> be the result of
    <a lt="collect documents to unfullscreen">collecting documents to unfullscreen</a> given


foolip:
This is also from https://codereview.chromium.org/2573773002/ where things just seemed to be complicated if it's possible to have no fullscreen element and trying to exit anyway, but I'm quite uncertain about the fix and would like feedback.

Another way would be to keep track of the original (potentially null) fullscreen element as *pending* and checking if it has changed, but I'd prefer not as exiting now doesn't need to keep track of any element at all. Together with #65, multiple calls to `exitFullscreen()` would at most behave as `exitFullscreen({ fully: true })` so maybe this is good enough.

fullscreen.bs
```diff
@@ -314,8 +314,8 @@ attribute's getter must run these steps:
 <ol>
  <li><p>Let <var>promise</var> be a new promise.
 
- <li><p>If <var>doc</var>'s <a>fullscreen element</a> is null, reject <var>promise</var> with a
- <code>TypeError</code> exception, and return <var>promise</var>.
+ <li><p>If <var>doc</var>'s <a>fullscreen element</a> is null, fulfill <var>promise</var> with
+ undefined, and return <var>promise</var>.
```
annevk: then resolve* and no comma before and. Same below.

annevk:
> The change to fulfill instead of reject is a drive-by for consistency.

Hmm, but we do reject in a number of places at the moment. So consistency with what?

fullscreen.bs
annevk: Ugh, I see I used "fulfill" in a couple places instead of "resolve". Do you want me to fix those first in a PR?

foolip: Will do!

foolip:
Actually, I found when looking at https://github.com/whatwg/fullscreen/pull/65 that these changes aren't entirely separable, so I will close this and make it part of that.

foolip:
> Hmm, but we do reject in a number of places at the moment. So consistency with what?

I had in mind the fact that `document.fullscreenElement.requestFullscreen()` will resolve with no events, at least if the fullscreen element ready check still passes. Let's see about this again in #65.

fullscreen.bs
annevk: Oh, thanks!

foolip:
Sorry @upsuper, I meant to add you do #65.

PR #68, opened 2016-12-15
Reject exitFullscreen() in inactive documents

### LINKED ISSUE #67: Define what happens with requestFullscreen() and exitFullscreen() in inactive document
From @zcorpan in https://bugs.chromium.org/p/chromium/issues/detail?id=674454

I don't think we should fire any events, because I don't know how to implement it. (In Blink, in an inactive document, events are just dropped on the ground.)

Just rejecting the promise would work, though.

Also need to define what happens if the document becomes inactive between the request and the animation frame task.

foolip:
If the document becomes inactive after the request, then there will not
be another animation frame task, in which case the promise will be left
hanging instead.

Fixes #67.

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -314,11 +314,9 @@ attribute's getter must run these steps:
 <ol>
  <li><p>Let <var>promise</var> be a new promise.
 
- <li><p>If <var>doc</var> is not <a>fully active</a>, then reject <var>promise</var> with a
- <code>TypeError</code> exception and return <var>promise</var>.
-
- <li><p>If <var>doc</var>'s <a>fullscreen element</a> is null, then reject <var>promise</var> with a
- <code>TypeError</code> exception and return <var>promise</var>.
+ <li><p>If <var>doc</var> is not <a>fully active</a> or <var>doc</var>'s <a>fullscreen element</a>
+ is null, then reject <var>promise</var> with a <code>TypeError</code> exception and return
+ <var>promise</var>.
 
  <li><p>Let <var>resize</var> be false.
 
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -314,8 +314,11 @@ attribute's getter must run these steps:
 <ol>
  <li><p>Let <var>promise</var> be a new promise.
 
- <li><p>If <var>doc</var>'s <a>fullscreen element</a> is null, reject <var>promise</var> with a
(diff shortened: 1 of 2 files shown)


fullscreen.bs
```diff
@@ -314,6 +314,10 @@ attribute's getter must run these steps:
 <ol>
  <li><p>Let <var>promise</var> be a new promise.
 
+ <li><p>If <var>doc</var> has no <a for=Document>browsing context</a>, or if <var>doc</var>'s
+ <a for=Document>browsing context</a>'s <a>active document</a> is not <var>doc</var>, then
```
zcorpan: I think this needs to be `if <var>doc</var> is not <a>fully active</a>`, since it might be the active document but an ancestor browsing context's document is not active.

foolip: Excellent, didn't know that concept existed.

annevk:
Do we have tests for this too?

fullscreen.bs
```diff
@@ -314,6 +314,9 @@ attribute's getter must run these steps:
 <ol>
  <li><p>Let <var>promise</var> be a new promise.
 
+ <li><p>If <var>doc</var> is not <a>fully active</a>, reject <var>promise</var> with a
```
annevk: then reject* (I realize it's not consistent, but I'd like new and changed text to conform to it at least.)

fullscreen.bs
```diff
@@ -314,6 +314,9 @@ attribute's getter must run these steps:
...
+ <code>TypeError</code> exception, and return <var>promise</var>.
```
annevk: No comma before "and". Hmm, maybe we should fix the next line as well.

foolip:
Style possibly fixed. Then I can learn, but I haven't been able to internalize the rules for commas, so we'll keep having fun with that :)

foolip:
I'll now write a test, which will be the first test fullscreen test involving a promise that will fail everywhere.

foolip:
Test is reviewed. @annevk, can you merge both together if all LGTY?

foolip:
@annevk, it looks like https://fullscreen.spec.whatwg.org/#exit-fullscreen hasn't been updated, are the auto-publish scripts misbehaving?

annevk:
Should be good now.

foolip:
Yep, thanks!

PR #69, opened 2016-12-16
Editorial: resolve promises, don't fulfill them

foolip:


foolip:
(If anything just seemed odd to you, it was because I noticed I missed to commit fullscreen.html and force pushed pretty much at the same time as you push the fix.)

annevk:
Yeah, I noticed when I tried to merge :-)

PR #72, opened 2017-01-06
Change algorithm of "collect documents to unfullscreen"

upsuper:
While reviewing #65, I noticed two issues in the current algorithm:
1. If the current document has more than a single fullscreen element, its browsering context container will be accidentally unfullscreened because in algorithm of "exit fullscreen", "doc" is added to "exitDocs" if "doc" has more than a single fullscreen element, and in that case, the document of the browsing context container of "doc" will be added to "exitDocs" in the next step.
2. A browsing context container can be unfullscreened accidentally by its child document even if its "iframe fullscreen flag" is set, because the steps in "exit fullscreen" doesn't check it.

This pull request changes the algorithm to collecting all documents which we are going to call "unfullscreen an element" on.

***
[Preview](https://s3.amazonaws.com/pr-preview/upsuper/fullscreen/collect-doc-unfullscreen.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fullscreen/8130c44..upsuper:collect-doc-unfullscreen:d8802c3.html)

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -301,13 +301,22 @@ consists of a single <a>element</a> that has its <a>fullscreen flag</a> set.
 <p>To <dfn>collect documents to unfullscreen</dfn> given <var>doc</var>, run these steps:
 
 <ol>
- <li><p>If <var>doc</var> is not a <a>simple fullscreen document</a>, return the empty set.
-
  <li><p>Let <var>docs</var> be an ordered set consisting of <var>doc</var>.
 
- <li><p>While <var>docs</var>'s last <a for=/>document</a> has a <a>browsing context container</a>
- whose <a>node document</a> is a <a>simple fullscreen document</a> and does not have its
- <a>iframe fullscreen flag</a> set (if any), append that <a>node document</a> to <var>docs</var>.
+ <li><p>Run these substeps until terminated:
+
+ <ol>
+  <li><p>Let <var>lastDoc</var> be <var>docs</var>'s last <a for=/>document</a>.
+
+  <li><p>If <var>lastDoc</var> is not a <a>simple fullscreen document</a>, terminate these substeps.
+
+  <li><p>Let <var>container</var> be <var>lastDoc</var>'s <a>browsing context container</a>
+  if any, and terminate these substeps otherwise.
+
+  <li><p>If <var>container</var>'s <a>iframe fullscreen flag</a> is set, terminate these substeps.
+
+  <li><p>Append <var>container</var>'s <a>node document</a> to <var>docs</var>.
+ </ol>
 
  <li><p>Return <var>docs</var>.
 </ol>
@@ -332,7 +341,8 @@ consists of a single <a>element</a> that has its <a>fullscreen flag</a> set.
  <a>active document</a>.
  <!-- cross-process -->
 
- <li><p>If <var>topLevelDoc</var> is in <var>docs</var>, then set <var>resize</var> to true.
+ <li><p>If <var>topLevelDoc</var> is in <var>docs</var>, and it is a
+ <a>simple fullscreen document</a>, then set <var>resize</var> to true.
 
  <li><p>Return <var>promise</var>, and run the remaining steps <a>in parallel</a>.
 
@@ -349,15 +359,10 @@ consists of a single <a>element</a> that has its <a>fullscreen flag</a> set.
    <var>doc</var>.
    <!-- cross-process -->
 
-   <li><p>If <var>resize</var> is true and <var>topLevelDoc</var> is not in <var>exitDocs</var>,
+   <li><p>If <var>resize</var> is true and <var>topLevelDoc</var> is either not in
+   <var>exitDocs</var>, or not a <a>simple fullscreen document</a>,
    <a>fully exit fullscreen</a> <var>topLevelDoc</var>, reject <var>promise</var> with a
    <code>TypeError</code> exception, and terminate these steps.
-
-   <li><p>If <var>exitDocs</var> is the empty set, append <var>doc</var> to <var>exitDocs</var>.
-
-   <li><p>If <var>exitDocs</var>'s last <a for=/>document</a> has a
-   <a>browsing context container</a>, append that <a>browsing context container</a>'s
-   <a>node document</a> to <var>exitDocs</var>.
    <!-- cross-process -->
 
    <li><p>Let <var>descendantDocs</var> be an ordered set consisting of <var>doc</var>'s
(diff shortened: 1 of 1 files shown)


upsuper:
#65 may need some change on top of this... or we can land #65 first than I rebase this one on top of that?

foolip:
@upsuper, I guess you still want to merge this? Can you rebase it so that there are no conflicts?

upsuper:
There wasn't realy a conflict. The only conflict was that the generated fullscreen.html was deleted in master.

fullscreen.bs
```diff
@@ -295,18 +295,28 @@ attribute's getter must run these steps:
...
- <li><p>If <var>doc</var>'s <a>top layer</a> consists of more than a single <a>element</a> that has
- its <a>fullscreen flag</a> set, return the empty set.
-
  <li><p>Let <var>docs</var> be an ordered set consisting of <var>doc</var>.
 
- <li><p>While <var>docs</var>'s last <a for=/>document</a> has a <a>browsing context container</a>
- whose <a>node document</a>'s <a>top layer</a> consists of a single <a>element</a> that has its
- <a>fullscreen flag</a> set and does not have its <a>iframe fullscreen flag</a> set (if any), append
- that <a>node document</a> to <var>docs</var>.
+ <li><p>Run these substeps until terminated:
```
foolip: Is this a `while(true)` loop? @annevk, any advice from https://infra.spec.whatwg.org/#algorithm-control-flow on how this should be phrased? I think this could plausibly be read as just running once through if not terminated.

fullscreen.bs
```diff
@@ -348,15 +359,10 @@ attribute's getter must run these steps:
    <var>doc</var>.
    <!-- cross-process -->
 
-   <li><p>If <var>resize</var> is true and <var>topLevelDoc</var> is not in <var>exitDocs</var>,
+   <li><p>If <var>resize</var> is true and <var>topLevelDoc</var> is either not in
+   <var>exitDocs</var>, or not a <a>simple fullscreen document</a>,
    <a>fully exit fullscreen</a> <var>topLevelDoc</var>, reject <var>promise</var> with a
```
foolip: I'm curious if Gecko actually has this "fully exit fullscreen" step when exiting? In Blink, at least after the (now reverted) attempt to align with the spec, this didn't make sense. This is the topic of https://github.com/whatwg/fullscreen/pull/65

I'm fine with this change if it's just incidental of course, and not the central point of this PR.

fullscreen.bs
annevk: Yeah, I might just say "While true:". Perhaps file an issue against Infra to introduce more loop constructs.

foolip:
Let me see if I understand the two problems here. Assuming that the top layers don't change during the transition:

> If the current document has more than a single fullscreen element, its browsering context container will be accidentally unfullscreened because in algorithm of "exit fullscreen", "doc" is added to "exitDocs" if "doc" has more than a single fullscreen element, and in that case, the document of the browsing context container of "doc" will be added to "exitDocs" in the next step.

https://fullscreen.spec.whatwg.org/#collect-documents-to-unfullscreen would return the empty set. Then "If exitDocs is the empty set, append doc to exitDocs" applies and the following step indeed adds the parent document to *exitDocs* as well.

This should be easy enough to write a test for to see the currently implemented behavior, and I don't see it covered in https://github.com/w3c/web-platform-tests/tree/master/fullscreen/api

> A browsing context container can be unfullscreened accidentally by its child document even if its "iframe fullscreen flag" is set, because the steps in "exit fullscreen" doesn't check it.

But the fix is still to just fix https://fullscreen.spec.whatwg.org/#collect-documents-to-unfullscreen and not repeat any "iframe fullscreen flag" outside of it, right? Just checking since that's what your changes do.

fullscreen.bs
```diff
@@ -295,18 +295,28 @@ attribute's getter must run these steps:
  <li><p>Return null.
 </ol>
 
+<p>A <a>document</a> is said to be a <dfn>simple fullscreen document</dfn> if its <a>top layer</a>
```
foolip: What do we think about documents with a single fullscreen element in the top layer but a dialog above or beneath it?

upsuper: That should still be a simple fullscreen document, since only one fullscreen element is in the top layer.

fullscreen.bs
upsuper: What we actually do in this case is that, after the window gets resized, we synchronously reset the whole document tree to non-fullscreen state regardless of whether the current state is identical to the previous state. Somehow it is simliar to this "fully exit fullscreen" step.

upsuper:
> This should be easy enough to write a test for to see the currently implemented behavior, and I don't see it covered in https://github.com/w3c/web-platform-tests/tree/master/fullscreen/api

Yep, that should be easy. Just have an iframe whose content page request fullscreen twice and exit once.

> But the fix is still to just fix https://fullscreen.spec.whatwg.org/#collect-documents-to-unfullscreen and not repeat any "iframe fullscreen flag" outside of it, right? Just checking since that's what your changes do.

Right, because in the new algorithm, outside the "collect documents to fullscreen", there would be nothing added to "exitDocs", and thus there is no need to check that flag outside it.

fullscreen.bs
foolip: Ah, so I would have read this definition as requiring that there is exactly one element in the top layer, and that it has the fullscreen flag set. Perhaps rephrase as "A document is said to be a simple fullscreen document if there is exactly one element in its top layer with the fullscreen flag set" and write a note saying that "In other words, a document with two elements in its top layer, where just one has the fullscreen flag set, is a simple fullscreen document"?

fullscreen.bs
foolip: "While true:" WFM, and I've filed https://github.com/whatwg/infra/issues/66

fullscreen.bs
foolip: This is what Blink does as well. I think this change is fine, I'll try to bring it closer to implementation in https://github.com/whatwg/fullscreen/pull/65 and ask for your review then.

foolip:
Tests for this (and a simpler nested case) are now up for review at https://codereview.chromium.org/2706293013/. I'll ask a Chromium reviewer, but if anyone wants to take a look and comment here, that's OK.

With the tests, there are just a few tweaks I asked for, so I'll make those now.

annevk:
Shit, I missed that you wanted @upsuper to review first. Hopefully all is good.

upsuper:
Yep, looks fine to me. Thanks for fixing that.

foolip:
Thanks all, I'll make sure those tests reach web-platform-tests soon.

PR #73, opened 2017-01-12
Meta: enable snapshots

annevk:


annevk:
@foolip @upsuper this PR basically changes nothing. The only problem is that once we land it existing PRs will need to be rebased and no longer include the HTML output. Let me know if that's a problem and I can hold off for a bit.

PR #75, opened 2017-02-06
Link to web-platform-tests in header

foolip:


foolip:
This will render as "<a href="https://github.com/w3c/web-platform-tests/tree/master/fullscreen">web-platform-tests/fullscreen</a> (<a href="https://github.com/w3c/web-platform-tests/issues?q=is%3Aopen+is%3Aissue+label%3Afullscreen">open issues</a>, <a href="https://github.com/w3c/web-platform-tests/pulls?q=is%3Aopen+is%3Apr+label%3Afullscreen">open PRs</a>)"

I think we should add something like this to all specs, and am starting here. In this case there are no open issues, while for dom there are plenty. Should we link to issues and PRs separately, or merge them as "open issues"?

If separate, does everyone know what "PRs" are, or would "open pull requests" be better?

foolip:
Credit where credit is due, seeing https://w3c.github.io/IndexedDB/ triggered this. @inexorabletash, have you given any thought to linking to open issues and PRs?

In various the context of test suite "ownership" has come up. I'd argue that it's just a convenience that web-platform-tests is all one repository, and that nobody can be responsible for triaging all issues and PRs. Instead, those who feel responsible of the spec (hopefully at least its maintainer and active implementers) might be the best people to worry about test suite health as well.

(Later, I'd also like to link to https://wptdashboard.appspot.com/ results. @jeffcarp)

jgraham:
I would like to see separate links to open issues and PRs; I think the audience for each is different. Open issues are mainly of interest to editors and those who might like to understand the status of and current problems in the spec. The set of people who might review PRs overlaps with this set, but not entirely; PR reviewers might be anyone who is familiar with the spec as written (e.g. for DOM I might review PRs without having any desire to wade through open issues).

I'm also slightly skeptical of linking the test results dashboard; there is a fine line to walk with that in making sure that implementors have access to it, and care about the results, without it becoming something that marketing care about. Prominent linking at the top of the spec seems too much like it might attract the latter kind of attention (which past experience suggests leads to bad outcomes in terms of people's willingness to submit tests they fail, or to not submit unnecessary redundant tests that the competition happens to fail).

foolip:
Very well, we can revisit the dashboard question on a per-spec basis once there is actually something to link to.

fullscreen.bs
```diff
@@ -13,6 +13,7 @@ Boilerplate: omit feedback-header, omit conformance, omit index, omit idl-index
 !Commits: <a href=https://github.com/whatwg/fullscreen/commits>GitHub whatwg/fullscreen/commits</a>
 !Commits: [SNAPSHOT-LINK]
 !Commits: <a href=https://twitter.com/fullscreenapi>@fullscreenapi</a>
+!Tests: <a href=https://github.com/w3c/web-platform-tests/tree/master/fullscreen>web-platform-tests/fullscreen</a> (<a href=https://github.com/w3c/web-platform-tests/issues?q=is%3Aopen+is%3Aissue+label%3Afullscreen>open issues</a>, <a href=https://github.com/w3c/web-platform-tests/pulls?q=is%3Aopen+is%3Apr+label%3Afullscreen>open PRs</a>)
```
annevk: Maybe "Web Platform Tests fullscreen/" as link text? As for the second and third link, I suggest we replace those with https://github.com/w3c/web-platform-tests/labels/fullscreen and something like "ongoing work" as text.

annevk:
Oh, I didn't review this thread before reviewing, my bad. I'm a little skeptical we'll have enough open issues/PRs to warrant two links and clicks for people that are just interested in what's going on. E.g., currently there are no open issues and I would expect those to be mostly filed against the standard since they're likely due to lack of interop.

If folks feel strongly I'm happy to go with separate links for now though. I don't think we can just use "PR". We should make it "pull request".

foolip:
I would be quite happy with a single link to https://github.com/w3c/web-platform-tests/labels/fullscreen and calling it simply "open issues" even though it includes PRs. "ongoing work" might not be so accurate for other specs where there are long-lived tracking bugs for missing test coverage and the like.

As for the link text, does "web-platform-tests fullscreen" work? Between web-platform-tests and Web Platform Tests, I have an affinity for the former, and I think @jgraham does too.

annevk:
"web-platform-tests fullscreen/" seems fine as well. You should also update the README as part of this PR to point out the tests. See https://github.com/whatwg/dom/blob/master/README.md#tests for example.

I don't like using "open issues" again since the link goes elsewhere. That does not seem great for users of assistive technology.

foolip:
Now updated.

README.md
```diff
@@ -40,4 +40,4 @@ If you can commit to this repository, see the
 ### Tests
 
 Tests can be found in the `fullscreen/` directory of
-[`w3c/web-platform-tests`](https://github.com/w3c/web-platform-tests).
+[`web-platform-tests`](https://github.com/w3c/web-platform-tests).
```
foolip: I wanted to do this as well, but thought it might have been deliberate in DOM :)

annevk: I think at some point @domenic convinced me to change this since it's a shared repository.

PR #78, opened 2017-05-04
Fix Bikeshed error about SVG reference

foolip:
Bikeshed complained that "[SVG] is replaced by [SVG11]."

Keep using [SVG] with no number, like for [CSS].

***
[Preview](https://fullscreen.spec.whatwg.org/branch-snapshots/svg-ref/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fullscreen/40c8947...c81af9d.html)

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -36,6 +36,9 @@ urlPrefix: https://w3c.github.io/screen-orientation/#dfn-
 {
     "CSS": {
         "aliasOf": "CSS2"
+    },
+    "SVG": {
+        "aliasOf": "SVG11"
     }
 }
 </pre>


annevk:
In HTML we use SVG11 as reference still. Should we update that too then?

foolip:
https://html.spec.whatwg.org/#dependencies says some pretty elaborate things about how to mix the various SVG specs, and what from SVG11 *shouldn't* be implemented. Untangling that would be nice, but presumably would require fixing some things in SVG2.

annevk:
In that case it seems better to keep using the SVG11 reference here.

foolip:
Reverted to SVG11.

annevk:
Ah, forgot to prefix with "Meta:". Next time!

PR #79, opened 2017-05-09
Meta: use infra for lists and sets

foolip:
This also clarifies the description of top layer order somewhat.

***
[Preview](https://fullscreen.spec.whatwg.org/branch-snapshots/usa-infra/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fullscreen/48fdc4d...5077749.html)

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -24,6 +24,8 @@ Boilerplate: omit feedback-header, omit conformance, omit index, omit idl-index
 spec:dom
     type:dfn; for:/; text:element
     type:interface; text:Document
+spec:infra
+    type:dfn; for:set; text:for each
 </pre>
 
 <pre class=anchors>
@@ -74,8 +76,8 @@ stated otherwise it is unset.
 and <a for="top layer">add</a> it to its <a>node document</a>'s <a>top layer</a>.
 
 <p>To <dfn>unfullscreen an <var>element</var></dfn>, unset <var>element</var>'s
-<a>fullscreen flag</a> and <a>iframe fullscreen flag</a> (if any), and <a for="top layer">remove</a>
-it from its <a>node document</a>'s <a>top layer</a>.
+<a>fullscreen flag</a> and <a>iframe fullscreen flag</a> (if any), and <a for=set>remove</a> it from
+its <a>node document</a>'s <a>top layer</a>.
 
 <p>To <dfn>unfullscreen a <var>document</var></dfn>,
 <a lt="unfullscreen an element">unfullscreen</a> all <a>elements</a>, within <var>document</var>'s
@@ -102,7 +104,7 @@ it from its <a>node document</a>'s <a>top layer</a>.
  that have their <a>fullscreen flag</a> set, in <a>shadow-including tree order</a>.
 
  <li>
-  <p>For each <var>node</var> in <var>nodes</var>, run these substeps:
+  <p><a>For each</a> <var>node</var> in <var>nodes</var>, run these substeps:
 
   <ol>
    <li><p>If <var>node</var> is its <a>node document</a>'s <a>fullscreen element</a>,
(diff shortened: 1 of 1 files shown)


fullscreen.bs
```diff
@@ -102,7 +105,7 @@ it from its <a>node document</a>'s <a>top layer</a>.
  that have their <a>fullscreen flag</a> set, in <a>shadow-including tree order</a>.
 
  <li>
-  <p>For each <var>node</var> in <var>nodes</var>, run these substeps:
+  <p><a>For each</a> <var>node</var> in <var>nodes</var>, run these substeps:
```
annevk: This needs a for attribute I think.

fullscreen.bs
```diff
@@ -229,18 +232,18 @@ these steps:
...
    <var>pending</var>.
 
    <li><p>While the first element in <var>fullscreenElements</var> is in a
-   <a>nested browsing context</a>, prepend its <a>browsing context container</a> to
+   <a>nested browsing context</a>, <a for=set>prepend</a> its <a>browsing context container</a> to
    <var>fullscreenElements</var>.
    <!-- cross-process -->
 
-   <li><p>Let <var>eventDocs</var> be an empty list.
+   <li><p>Let <var>eventDocs</var> be an empty <a>list</a>.
```
annevk: This can be a set as well?

fullscreen.bs
```diff
@@ -229,18 +232,18 @@ these steps:
...
 
    <li>
-    <p>For each <var>element</var> in <var>fullscreenElements</var>, in order, run these
+    <p><a>For each</a> <var>element</var> in <var>fullscreenElements</var>, run these
```
annevk: Needs a for attribute I think.

fullscreen.bs
```diff
@@ -252,16 +255,16 @@ these steps:
 
       <p class=note>No need to notify observers when nothing has changed.
 
-     <li><p>Otherwise, append <var>doc</var> to <var>eventDocs</var>.
+     <li><p>Otherwise, <a for=list>append</a> <var>doc</var> to <var>eventDocs</var>.
```
annevk: If we change eventDocs to be a set, this needs to change.

fullscreen.bs
```diff
@@ -252,16 +255,16 @@ these steps:
...
 
      <li><p>If <var>element</var> is <var>pending</var> and <var>pending</var> is an <{iframe}>
      <a>element</a>, then set <var>element</var>'s <a>iframe fullscreen flag</a>.
 
      <li><p><a lt="Fullscreen an element">Fullscreen <var>element</var></a> within <var>doc</var>.
     </ol>
 
-   <li><p>For each <var>doc</var> in <var>eventDocs</var>, in order, <a>fire an event</a> named
-   <code>fullscreenchange</code> on <var>doc</var>.
+   <li><p><a>For each</a> <var>doc</var> in <var>eventDocs</var>, <a>fire an event</a>
```
annevk: Needs a for attribute.

annevk: And a couple more of these below.

fullscreen.bs
```diff
@@ -496,11 +499,15 @@ following characteristics:
...
+<p>To <dfn export for="top layer">add</dfn> an <var>element</var> to a <var>top layer</var>,
+<a for=set>remove</a> it from <var>top layer</var> and then <a for=set>append</a> it to
+<var>top layer</var>.
+
+<p class=note>In other words, <var>element</var> is moved to the end of <var>top layer</var> if it
+is already present.
 
 <p>To <dfn export for="top layer">remove</dfn> an <var>element</var> from a <var>top layer</var>,
-remove <var>element</var> from <var>top layer</var>.
+<a for=set>remove</a> <var>element</var> from <var>top layer</var>.
```
annevk: Given that these are identical, perhaps we should no longer have a dedicated remove entry point?

foolip:
> A couple things around documents here will change though right if we change the setup in HTML? But maybe you're making this cleanup change first so that goes more smoothly.

Yeah, some of this will go away, some for loops will move, so I figured it's best to avoid those discussions in the real upcoming changes.

fullscreen.bs
foolip: Sure. I'm going to suggest removing it next, but let's not assume that's the final shape of the spec, more compat problems might revert it.

fullscreen.bs
foolip: Since only "for each" is used now, I've gone with "type:dfn; for:set; text:for each" to make the source a bit more readable.

fullscreen.bs
foolip: Doing that too.

fullscreen.bs
foolip: Sure, let's say that the top layer *is* an ordered set instead of that it *consists* of it, that'll make it natural to remove the extra remove layer.

foolip: Ugh, but HTML uses #top-layer-remove. Now I'm going to pretend that it's deliberate that top layer is an abstraction layer you have to go through in case we want to change what kind of data structure it is :)

foolip:
OK, https://github.com/whatwg/fullscreen/pull/79 and https://github.com/whatwg/html/pull/2650 good to go I think, but don't assume I didn't make mistakes :)

PR #81, opened 2017-05-10
Clarify what "collect documents to unfullscreen" returns

foolip:
Given that there is an "unfullscreen a document" operation, this
algorithm can easily be assumed to be the set of documents for which
that should be called, i.e., that all of the documents returned will
exit fully.

At the beginning of the animation frame task there's no guarantee that
doc still has a fullscreen element, but later in "unfullscreen exitDoc's
fullscreen element" it is assumed to have one. Fix this with an early
return.

***
[Preview](https://fullscreen.spec.whatwg.org/branch-snapshots/collect-documents-to-unfullscreen/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fullscreen/c8164af...ae151d8.html)

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -320,6 +320,8 @@ could be an open <{dialog}> element.
   <ol>
    <li><p>Let <var>lastDoc</var> be <var>docs</var>'s last <a for=/>document</a>.
 
+   <li><p>Assert: <var>lastDoc</var>'s <a>fullscreen element</a> is not null.
+
    <li><p>If <var>lastDoc</var> is not a <a>simple fullscreen document</a>, <a>break</a>.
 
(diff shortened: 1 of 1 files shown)


foolip:
Updated the commit description with a pointer to https://github.com/w3c/web-platform-tests/pull/5872

annevk:
LGTM. Will let @upsuper review as well.

upsuper:
LGTM

foolip:
Volunteers to review https://github.com/w3c/web-platform-tests/pull/5872?

I won't be filing any browser bugs, because nobody supports promises yet.

foolip:
(I merged this first so that I can rebase https://github.com/whatwg/fullscreen/pull/65 and because I'm confident the test change will be fine.)

PR #83, opened 2017-05-11
Editorial: use Infra's break to clarify requestFullscreen() iteration

foolip:
The implementation in Blink uses continue here, which would matter if
one of the ancestor documents had a fullscreen element other than the
iframe containing the document now requesting fullscreen.

Perhaps this should be continue, but that would be a separate change.

***
[Preview](https://fullscreen.spec.whatwg.org/branch-snapshots/infra-break/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fullscreen/0a4787a...15e285f.html)

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -243,15 +243,13 @@ these steps:
    <li><p>Let <var>eventDocs</var> be a new <a>ordered set</a>.
 
    <li>
-    <p><a>For each</a> <var>element</var> in <var>fullscreenElements</var>, run these
-    subsubsteps:
+    <p><a>For each</a> <var>element</var> in <var>fullscreenElements</var>:
 
     <ol>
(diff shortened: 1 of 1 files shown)


annevk:
I actually think the original meaning would be continue. Since you only abort the subsubsteps, not the for each, which would be the substep.

annevk:
(Good showcase for why the whole substep approach was misguided.)

foolip:
If that is how you would read the original spec, I'd be happy to change it to that instead, since it matches Blink. I just didn't want to commit myself to writing a test right now, so I need that "Editorial:" prefix :)

annevk:
Yeah, that's how I'd read it and I'd be fine with it being editorial.

foolip:
Done.

PR #85, opened 2017-05-12
Require that the document is fully active in requestFullscreen()

foolip:
Test: https://github.com/w3c/web-platform-tests/pull/5901

***
[Preview](https://fullscreen.spec.whatwg.org/branch-snapshots/request-fully-active/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fullscreen/3513c94...b1b684a.html)

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -193,9 +193,14 @@ these steps:
 <ol>
  <li><p>Let <var>pending</var> be the <a>context object</a>.
 
- <li><p>Let <var>error</var> be false.
+ <li><p>Let <var>pendingDoc</var> be <var>pending</var>'s <a>node document</a>.
 
  <li><p>Let <var>promise</var> be a new promise.
+
+ <li><p>If <var>pendingDoc</var> is not <a>fully active</a>, then reject <var>promise</var> with a
+ <code>TypeError</code> exception and return <var>promise</var>.
+
+ <li><p>Let <var>error</var> be false.
 
  <li>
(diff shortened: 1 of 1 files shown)


foolip:
This is similar to the check in exitFullscreen(), and Blink already has a check like this, which is why I'd like to get it into the spec. Firefox actually fails the test in https://github.com/w3c/web-platform-tests/pull/5901 because it fires a fullscreenerror event, a behavior I think would be hard to implement in Blink (trying to fire an event in an inactive document does nothing last I checked) and isn't super useful.

fullscreen.bs
```diff
@@ -193,10 +193,15 @@ these steps:
 <ol>
  <li><p>Let <var>pending</var> be the <a>context object</a>.
 
- <li><p>Let <var>error</var> be false.
+ <li><p>Let <var>pendingDoc</var> be <var>pending</var>'s <a>node document</a>.
 
  <li><p>Let <var>promise</var> be a new promise.
 
+ <li><p>If <var>pendingDoc</var> is not <a>fully active</a>, then reject <var>promise</var> with a
+ <code>TypeError</code> exception and return <var>promise</var>.
```
annevk: You cannot return here.

foolip: Why not? This is before the "in parallel" steps.

annevk: I wrongly assumed that we need to run those either way.

PR #86, opened 2017-05-12
Animation frame tasks

foolip:


foolip:
I opened and closed this to have something to point to in https://github.com/whatwg/html/pull/2627 while I continue to work out the changes that should precede this, and making sure that this really matches what I have implemented or plan to implement.

PR #87, opened 2017-05-16
In requestFullscreen(), handle the pending element moving

foolip:
Fixes https://github.com/whatwg/fullscreen/issues/33.

Drive-by: capitalization in lt attributes to match dfn.

***
[Preview](https://fullscreen.spec.whatwg.org/branch-snapshots/pending-element-moved/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fullscreen/d1d1b4a...e96e446.html)

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -90,7 +90,7 @@ its <a>node document</a>'s <a>top layer</a>.
 <ol>
  <li><p>If <var>document</var>'s <a>fullscreen element</a> is null, terminate these steps.
 
- <li><p><a lt="Unfullscreen an element">Unfullscreen elements</a> whose <a>fullscreen flag</a> is
+ <li><p><a lt="unfullscreen an element">Unfullscreen elements</a> whose <a>fullscreen flag</a> is
  set, within <var>document</var>'s <a>top layer</a>, except for <var>document</var>'s
  <a>fullscreen element</a>.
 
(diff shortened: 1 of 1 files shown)


foolip:
A test that already tests this was added in https://github.com/w3c/web-platform-tests/pull/4250:
http://w3c-test.org/fullscreen/api/element-request-fullscreen-and-move-to-iframe-manual.html

Firefox Nightly already passes this test.

It was an oversight that this test was included before the spec change, I've gone through the PR and checked for other things not yet reflected in the spec, but can't spot anything.

foolip:
Merged, but can you confirm if this will work for Gecko, @upsuper? Maybe the tests doesn't exercise some minor difference between the spec and what Gecko does.

upsuper:
There is an issue here that, it seems when this happens, the browser window would end up being in fullscreen state while the content is not? It doesn't sound like a helpful state, but this is probably a separate issue.

Anyway, I think it works for Gecko, and I think except having the issue above, Gecko should have been doing something like this now, but I'm not completely sure.

foolip:
Yeah, that is a problem, and IIRC you have a fix for it in Gecko? I've filed https://github.com/whatwg/fullscreen/issues/88 but probably won't get back to it myself anytime soon since it's such a corner case.

upsuper:
Gecko doesn't have this problem. The algorithm is something like, when we are applying fullscreen state to the document because we just finish changing the viewport, if there is nothing successfully get fullscreened eventually, we revert the window state. I don't know if something like this is specable...

PR #91, opened 2017-05-22
Reinstate the hierarchy restrictions

foolip:
This reinstates parts of the fullscreen element ready check removed in
commit 766dc872bcf26c012549f7bdfde21c46bebba40c and
commit 9592913bbeca3a9d029274f310eb27ad8003622c, with some
simplification of the iframe ancestor check.

***
[Preview](https://fullscreen.spec.whatwg.org/branch-snapshots/hierarchy-restrictions/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fullscreen/cbdc4f7...fb22071.html)

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -182,6 +182,12 @@ if all of the following are true, and false otherwise:
 <ul>
  <li><p><var>element</var> is <a>connected</a>.
 
+ <li><p><var>element</var>'s <a>node document</a>'s <a>fullscreen element</a> is null, or is an
+ <a>inclusive ancestor</a> of <var>element</var>.
+ <!-- If /element/ is the fullscreen element, nothing happens -->
+
+ <li><p><var>element</var> has no <a>ancestor</a> <{iframe}> <a>element</a>.
+
  <li><p><var>element</var>'s <a>node document</a> is <a>allowed to use</a> the feature indicated by
  attribute name <code>allowfullscreen</code>.
  <!-- cross-process, recursive -->
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -223,6 +223,8 @@ these steps:
    <a href=https://www.w3.org/Math/draft-spec/chapter2.html#interf.toplevel>MathML <code>math</code></a>
    element. [[!SVG]] [[!MATHML]]
 
+   <li><p><var>pending</var> is not a <{dialog}> <a>element</a>.
+
    <li><p>The <a>fullscreen element ready check</a> for <var>pending</var> returns true.
 
    <li><p><a>Fullscreen is supported</a>.
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -191,6 +191,11 @@ if all of the following are true, and false otherwise:
  <li><p><var>element</var>'s <a>node document</a> is <a>allowed to use</a> the feature indicated by
  attribute name <code>allowfullscreen</code>.
  <!-- cross-process, recursive -->
+
+ <li><p><var>element</var>'s <a>node document</a>'s <a>browsing context</a> either has a
+ <a>browsing context container</a> and the <a>fullscreen element ready check</a> returns true for
+ that <a>browsing context container</a>, or it has no <a>browsing context container</a>.
+ <!-- cross-process, recursive -->
 </ul>
 
 <p>The <dfn method for=Element><code>requestFullscreen()</code></dfn> method, when invoked, must run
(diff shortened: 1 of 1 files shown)


foolip:
@upsuper, in addition to these changes, something needs to be done to uphold the invariant when mixed with `dialog`. I see two options:
* Add checks in HTML to avoid moving elements within top layer if it would end up violating the hierarchy restrictions.
* Add something in https://fullscreen.spec.whatwg.org/#top-layer-add to remove elements to maintain the invariants.

What does Gecko do?

upsuper:
Sadly, Gecko still doesn't have `dialog`... so this is not applicable.

I don't recall exactly what the problems are for that...

It seems one of the problem would be that `showModal` may be called on a dialog element which is already a fullscreen element?

For that, I would probably prefer the second way, specifically, we pop out fullscreen elements until the dialog is no longer a fullscreen element, and then we add the element back as a dialog.

But thinking about this a bit further, this option is probably unnecessarily complicated, because that means we would need to consider dispatching event, and even unfullscreen the window, for a `showModal` call, which sounds unfortunate.

Given this, I guess we should make HTML avoid invoking top-layer-add when the element is already a fullscreen element by rejecting it with a exception, and I think we should assert in top-layer-add that the element to be append is not currently an element in the top layer.

If we add that assertion, we would also need to think about what if a modal dialog requests fullscreen. Maybe we should reject as well. I don't think those cases are common, and I don't have idea how authors may find those cases useful...

fullscreen.bs
```diff
@@ -182,6 +182,12 @@ if all of the following are true, and false otherwise:
 <ul>
  <li><p><var>element</var> is <a>connected</a>.
 
+ <li><p><var>element</var>'s <a>node document</a>'s <a>fullscreen element</a> is null, or is an
+ <a>inclusive ancestor</a> of <var>element</var>.
```
upsuper: You also need to invoke this check recursively for its browsing context container if any.

foolip: Oh, right, the ready check used to be recursive but I removed that in https://github.com/whatwg/fullscreen/pull/52.

foolip:
>  I guess we should make HTML avoid invoking top-layer-add when the element is already a fullscreen element by rejecting it with a exception

Yep, we could add another "InvalidStateError" case to https://html.spec.whatwg.org/#dom-dialog-showmodal

> If we add that assertion, we would also need to think about what if a modal dialog requests fullscreen. Maybe we should reject as well.

Actually, if we add a check for `dialog` in `requestFullscreen()`, then no check should be needed in `showModal()`. With the current state of things, no `dialog` elements in top layer would be fullscreen, and all non-`dialog` elements would be.

upsuper:
> Actually, if we add a check for dialog in requestFullscreen(), then no check should be needed in showModal(). With the current state of things, no dialog elements in top layer would be fullscreen, and all non-dialog elements would be.

I mean we check whether the element is a modal dialog, not that whether it is a dialog element. But well, I guess we can just simple reject dialog from entering fullscreen if that sounds simpler... but could that be breaking?

foolip:
It could be, I'll add use counters in Blink both for when the element is a `dialog` element and when it is additionally in the top layer, which would mean that it's showing as modal. I think it's fairly unlikely that this would break real content, but we could leave it as an open issue until the results are in, or make a tentative change and revert if necessary.

foolip:
https://chromium-review.googlesource.com/c/512163/

foolip:
OK, I've pushed something that should work. I worry a bit about making the ready check recursive again, because it's a cross-process access and can't actually be checked synchronously. In other words, hierarchy restrictions is a simplifications in some ways, but not in others. However, I lean towards having them for the time being.

@upsuper, are you OK with the `dialog` bit, or should I add a TODO there and file a separate issue, blocked on usage data? Yet another possibility would be to change https://fullscreen.spec.whatwg.org/#top-layer-add to do nothing if the element already exists, so that things can never move in the top layer. Then `showModal()` would need to handle that situation instead.

upsuper:
I think this is fine. I'd still want to also change the "top-layer-add" algorithm to not move things around, but instead assert that the element to add is not in the top layer currently. It seems to me the HTML spec doesn't allow a open dialog to be opened again, so it wouldn't break that assertion either.

foolip:
Yeah, I think you're right that such an assert would hold without any other changes. I've pushed the assert you suggested, and think that https://github.com/whatwg/html/pull/2341 might need a tweak for that.

foolip:
Abandoning this.

foolip:
Sorry, I moved a bit fast when abandoning old PRs after https://github.com/whatwg/fullscreen/pull/128 obsoleted a lot of them. I'm a bit swamped until mid-next-week, but will come around to https://github.com/whatwg/fullscreen/issues/140.

Issue #74, opened 2017-01-26
document.fullscreenElement in scroll event after entering fullscreen matters for compat


foolip:
In https://crbug.com/680467 there is fallout from aligning the timing of `document.fullscreenElement` and `fullscreenchange` event with animation frames in https://crbug.com/402376.

On https://twitter.com/BCCI, if ones scrolls down a bit and plays a video in fullscreen, the iframe the video is in is removed from the document, leaving things in a broken-looking state. I [speculate](https://bugs.chromium.org/p/chromium/issues/detail?id=680467#c18) that this code tries to remove video that have scrolled out of view.

The crucial change is that in the scroll event fired after entering fullscreen, the fullscreen steps have not yet run, so to the page it looks like the page has simply been resized, causing a scroll.

I [investigated](https://bugs.chromium.org/p/chromium/issues/detail?id=680467#c20) current behavior:
* In Edge 14, the scroll event is fired after the fullscreenchange event, and so naturally the fullscreen element has already changed.
* In Firefox 54, the scroll event is fired before the fullscreenchange event when the fullscreen element has not yet changed.  For some reason, it still does not get this bug on Twitter.
* In Chrome 55 (before my changes) and Safari 10, the scroll event is fired before the fullscreenchange event, and yet the fullscreen element has already changed.

I can see a few alternatives here, in rough order of preference:
* Apply the fullscreen state changes immediately on resize and delay only the events until animation frame timing.
* Tweak the order of scroll and fullscreen change events in https://html.spec.whatwg.org/multipage/webappapis.html#event-loop-processing-model
* Introduce a `document.pendingFullscreenElement` that could be used by Twitter to avoid this problem.
* Ask Twitter to keep track of whether they are entering fullscreen themselves to avoid this.
* Suppress scroll events in fullscreen entirely.

@upsuper @jernoble @aliams

upsuper:
> Apply the fullscreen state changes immediately on resize and delay only the events until animation frame timing.

FWIW, this is what Gecko does, actually. I thought it is hard to distinguish between this behavior and the specced behavior. It seems I was wrong.

upsuper:
BTW, if the element is already removed from the document, I suppose browser should exit fullscreen anyway... unless they just hide the element.

foolip:
[...]

PR #92, opened 2017-05-31
Fullscreen/unfullscreen ASAP; fire events at animation frame timing

foolip:
This means that changes to document.fullscreenElement and other state
will be observable as soon as the resize itself is (e.g. via
window.innerWidth) and before resize or scroll events are fired.

The fullscreenchange event is still delayed to animation frame timing.

This also includes a slight change when /resize/ is true in "exit
fullscreen". By changing /doc/ to /topLevelDoc/ in this case, we can
make sure that we always fully unfullscreen all documents in this case.
This makes a "fully exit fullscreen" corner case unnecessary.

Fixes https://github.com/whatwg/fullscreen/issues/74.

***
[Preview](https://fullscreen.spec.whatwg.org/branch-snapshots/fullscreen-asap/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fullscreen/aaa10c1...a62b790.html)

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -26,6 +26,7 @@ spec:dom
     type:interface; text:Document
 spec:infra
     type:dfn; for:set; text:for each
+    type:dfn; text:string
 </pre>
 
 <pre class=anchors>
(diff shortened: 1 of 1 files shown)


foolip:
This depends on https://github.com/whatwg/html/pull/2627. @upsuper, would you mind reviewing both? I'm fine with not making animation frame tasks a generic concept if you think that's better, it wouldn't affect the important changes here.

foolip:
Ping @upsuper.

upsuper:
Sorry for the delay. Please give me another day. I'll try to review it tomorrow.

upsuper:
And yeah, I'd probably prefer we don't make animation frame tasks a generic concept. Having each thing in its own queue makes order of them clearer, and would also allow us to specify the order if we need / want, at least for this kind of things which can potentially affect rendering.

foolip:
Thanks for the review, @upsuper. I've now pushed an update that defines the "run the fullscreen rendering steps" already references by HTML, which means this doesn't depend on any HTML changes.

The implementation I hope to land in Blink is more similar to the "animation frame tasks" model I had, but currently it is indistinguishable. I might want to revisit https://github.com/whatwg/html/pull/2627 later, but it's not important today.

Care to take another look? If it's OK with you, I plan to write the tests for this in https://chromium-review.googlesource.com/521162, so you wouldn't need to do that review.

foolip:
To repeat, tests will be added in https://chromium-review.googlesource.com/521162

PR #100, opened 2017-08-30
Meta: update the URL for the common deploy.sh script

foolip:
Part of https://github.com/whatwg/whatwg.org/issues/71.

Also require the dist to be Trusty.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: trusty
 language: generic
 
 env:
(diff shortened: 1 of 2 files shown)


foolip:
> Maybe we should make .travis invoke make to avoid having to change two locations going forward?

Done, and I'll make the same change everywhere. I'm thinking we should merge all of these changes at the same time, in case later reviews reveals something we should change everywhere.

annevk:
That's fine. I'm happy for you to push these to master by the way when the repository doesn't require review. Provided you're reasonably sure it works.

foolip:
OK, so once @domenic has had a chance to take a look at the general shape of this, I'll prepare the remaining PRs, get review only for those that require it, and then merge them one by one.

PR #126, opened 2018-05-16
Adjust algorithm for FullyExitFullscreen.

dtapuska:
To exit fullscreen we exit all the documents that have fullscreen elements
posting messages to them and then resize the window to normal dimensions.

***
<a href="https://whatpr.org/fullscreen-1/126.html" title="Last updated on May 16, 2018, 8:44 PM GMT (50f2677)">Preview</a> | <a href="https://whatpr.org/fullscreen/126/8bbb8a0...50f2677.html" title="Last updated on May 16, 2018, 8:44 PM GMT (50f2677)">Diff</a>

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -79,33 +79,32 @@ its <a>node document</a>'s <a>top layer</a>.
 <p>To <dfn>fully exit fullscreen</dfn> a <a for=/>document</a> <var>document</var>, run these steps:
 
 <ol>
- <li><p>If <var>document</var>'s <a>fullscreen element</a> is null, terminate these steps.
+ <li><p>Let <var>topLevelDoc</var> be <var>document</var>'s <a>top-level browsing context</a>'s
+ <a>active document</a>.
 
- <li><p><a lt="unfullscreen an element">Unfullscreen elements</a> whose <a>fullscreen flag</a> is
- set, within <var>document</var>'s <a>top layer</a>, except for <var>document</var>'s
- <a>fullscreen element</a>.
+ <li><p>If <var>topLevelDoc</var>'s <a>fullscreen element</a> is null, terminate these steps.
 
- <li><p><a>Exit fullscreen</a> <var>document</var>.
-</ol>
+ <li><p>Let <var>exitDocs</var> be an <a>ordered set</a> consisting of <var>doc</var>.
 
-<p>Whenever the <a>removing steps</a> run with a <var>removedNode</var>, run these steps:
-
-<ol>
- <li><p>Let <var>nodes</var> be <var>removedNode</var>'s
- <a>shadow-including inclusive descendants</a> that have their <a>fullscreen flag</a> set, in
- <a>shadow-including tree order</a>.
+ <li><p>Append <var>topLevelDoc</var>'s <a>descendant browsing contexts</a>' <a>active documents</a>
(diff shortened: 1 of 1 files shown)


upsuper:
Is this the change I proposed in #102 comment?

dtapuska:
Yes it is. Although in your code you didn't append the current doc to the set of docs. So it is slightly adjusted. 

@foolip indicated that ordering doesn't matter between fullscreenchange and resize on a chat with him so I think this approach is easier to understand.

fullscreen.bs
```diff
@@ -79,34 +79,33 @@ its <a>node document</a>'s <a>top layer</a>.
...
-   <li><p>Otherwise, <a lt="unfullscreen an element">unfullscreen <var>node</var></a> within its
-   <a>node document</a>.
+   <li><p><a lt="unfullscreen a document">Unfullscreen <var>exitDoc</var></a>.
   </ol>
+
+ <li><p>Run the remaining steps <a>in parallel</a>.
+ <li><p>Resize <var>topLevelDoc</var>'s viewport to its "normal" dimensions.
 </ol>
 
+<p>Whenever the <a>removing steps</a> run with a <var>removedNode</var>,
```
foolip: This will unconditionally run "fully exit fullscreen" when any node is removed. https://fullscreen.spec.whatwg.org/#removing-steps (here removed) will need to retain at least the bits up to "For each node in nodes" to only deal with elements "that have their fullscreen flag set".

foolip:
OK, so treating this as part of the solution for https://github.com/whatwg/fullscreen/pull/102, the comment threads to resolve are https://github.com/whatwg/fullscreen/pull/102#pullrequestreview-92143962 and https://github.com/whatwg/fullscreen/pull/102#pullrequestreview-93144205.

fullscreen.bs
```diff
@@ -79,34 +79,33 @@ its <a>node document</a>'s <a>top layer</a>.
...
 
 <ol>
- <li><p>If <var>document</var>'s <a>fullscreen element</a> is null, terminate these steps.
+ <li><p>Let <var>topLevelDoc</var> be <var>document</var>'s <a>top-level browsing context</a>'s
+ <a>active document</a>.
 
- <li><p><a lt="unfullscreen an element">Unfullscreen elements</a> whose <a>fullscreen flag</a> is
- set, within <var>document</var>'s <a>top layer</a>, except for <var>document</var>'s
- <a>fullscreen element</a>.
+ <li><p>If <var>topLevelDoc</var>'s <a>fullscreen element</a> is null, terminate these steps.
```
foolip: How should this interact with the changes in https://github.com/whatwg/fullscreen/pull/102? The "fullscreen element is null" condition should be impossible if top layer is modified *after* the synchronous part of this algorithm has run.

fullscreen.bs
upsuper: We just need to check whether there exists any element in the subtree which has fullscreen flag set for this change.

fullscreen.bs
upsuper: The condition is possible because there are other callsites to this algorithm, for example, unloading document, and probably for the escaping mechanism. If we don't check here, we would need to guard all callsites with this condition, and we probably still want to assert it in this algorithm to ensure everything runs as expected.

foolip:
Superseded by https://github.com/whatwg/fullscreen/pull/128

Issue #123, opened 2018-03-27
Request Fullscreen take additional options


dtapuska:
There are multiple uses of fullscreen, videos, gaming, etc. Sometimes an app may wish to just be transient full screen and want to just hide the status and URL bar.

Does it make sense to augment requestFullScreen with a dictionary so that applications can possibly not set 

This can be done in native apps on Android by not specifying [SYSTEM_UI_FLAG_HIDE_NAVIGATION](https://developer.android.com/reference/android/view/View.html#SYSTEM_UI_FLAG_HIDE_NAVIGATION)

As well iOS has some APIs such as [hidesBottomBarWhenPushed](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621863-hidesbottombarwhenpushed) and [prefersstatusbarhidden](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621440-prefersstatusbarhidden)

annevk:
Maybe, if we do anything here we should reconcile it a bit with `window.open()` perhaps.

upsuper:
I don't think I fully understand what you want here.

Fullscreen means making the content take the whole screen area for it to display, and having UI hidden is an requirement for that to happen. But it seems you are proposing just hide the UI stuff?

Native apps on mobile system are kind of different, because those apps almost always already take the majority of the viewport, and thus hiding the system bars really means it's taking the whole screen. However, on desktop platforms, there are browser UI, as well as system UI, and browser may not take the whole desktop area from the beginning (when they are not maximized).

So it is unclear to me what you really want. from that analogy

dtapuska:
A page may wish to go fullscreen but allow an app to go back easily. On devices with physical buttons for back/home it is easy but on pages that have overlays for that it may be harder for the user to detect how to go back. So I'm proposing a property like `prefersNavigationBar` which would not set the HIDE_NAVIGATION flag.

Samsung's browser always does this; it never hides the navigation bar however Chrome does. So I'm not certain how we can reconcile the default.

winstonchen:
@dtapuska In response to your comment, "Samsung's browser always does this; it never hides the navigation bar...", Samsung Internet auto hides the navigation in some scenarios (ie. videos). 

@foolip We will get back to you on your questions.

dtapuska:
@winstonchen  Yes I saw this and should have been more descriptive. Fundamentally the webpage has the best knowledge of whether it is a immersive experience or not and that is what this request for options is about.

[...]

PR #129, opened 2018-06-18
Add FullscreenOptions dictionary to requestFullscreen

dtapuska:
Add an attribute of the FullscreenOptions dictionary that hints
at whether a navigation bar is preferred or not.

***
<a href="https://whatpr.org/fullscreen-1/129.html" title="Last updated on Sep 7, 2018, 7:12 PM GMT (af61df4)">Preview</a> | <a href="https://whatpr.org/fullscreen/129/06c6e07...af61df4.html" title="Last updated on Sep 7, 2018, 7:12 PM GMT (af61df4)">Diff</a>

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -273,15 +273,34 @@ when invoked, must run these steps:
 
  <li>
   <p>If <var>error</var> is false: Resize <var>pendingDoc</var>'s
-  <a>top-level browsing context</a>'s <a>active document</a>'s viewport's dimensions to match the
-  dimensions of the available screen of the output device, optionally taking into account
-  <var>options</var>'s <code>navigationUI</code> member. Optionally display a message how the end
-  user can revert this.
+  <a>top-level browsing context</a>'s <a>active document</a>'s viewport's dimensions, optionally
+  taking into account <var>options</var>'s <code>navigationUI</code> member:
   <!-- cross-process -->
 
-  <p class=note>The <code>navigationUI</code> member is a hint to the user agent about what the
-  author thinks will lead to the best user experience. The member might be ignored altogether, for
-  example based on user preferences.
+  <table>
+    <thead>
+      <tr>
+      <th>value</th>
+      <th>viewport dimensions</th>
+      </tr>
+    </thead>
+    <tbody>
+      <tr>
+        <td>"<code>hide</code>"</td>
+        <td>full dimensions of the screen of the output device</td>
+      </tr>
+      <tr>
+        <td>"<code>show</code>"</td>
+        <td>dimensions of the screen of the output device clamped to allow the user agent to show page navigation controls</td>
+      </tr>
+      <tr>
+        <td>"<code>auto</code>"</td>
+        <td>user-agent defined, but matching one of the above</td>
+      </tr>
+    </tbody>
+  </table>
+
+  <p>Optionally display a message how the end user can revert this.
 
  <li>
   <p>If any of the following conditions are false, then set <var>error</var> to true:
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -199,8 +199,15 @@ partial interface DocumentOrShadowRoot {
      interface, which is implemented by Document and HTMLElement, not Element. -->
 
 <dl class=domintro>
- <dt><code><var>promise</var> = <var>element</var> . {{Element/requestFullscreen(<var>options</var>)}}</code>
- <dd><p>Displays <var>element</var> fullscreen with <var>options</var> and resolves <var>promise</var> when done.
+ <dt><code><var>promise</var> = <var>element</var> . <a method for=Element lt=requestFullscreen()>requestFullscreen([<var>options</var>])</a></code>
+ <dd>
+  Displays <var>element</var> fullscreen and resolves <var>promise</var> when done.
+
+  When supplied, <var>options</var>'s <code>navigationUI</code> member indicates whether showing
+  navigation UI while in fullscreen is preferred or not. If set to "<code>show</code>", navigation
+  simplicity is preferred over screen space, and if set to "<code>hide</code>", more screen space
+  is preferred. User agents are always free to honor user preference over the application. The
+  default value "<code>auto</code>" indicates no application preference.
 
  <dt><code><var>document</var> . {{Document/fullscreenEnabled}}</code>
  <dd><p>Returns true if <var>document</var> has the ability to display <a>elements</a> fullscreen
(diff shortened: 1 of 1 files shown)


dtapuska:
@foolip  Please read and let me know what I should change. Once it is ok; I'll put together wpt before landing.

fullscreen.bs
```diff
@@ -155,8 +155,12 @@ steps:
 <h2 id=api>API</h2>
 
 <pre class=idl>
+dictionary <dfn>FullscreenOptions</dfn> {
+  boolean prefersNavigationBar = true;
```
foolip: As a rule of thumb, to which there are only a few exceptions, optional boolean dictionary members should default to false. This is because undefined is a falsy value, but treated the same as missing, which means that `{ prefersNavigationBar: undefined }` behaves like `{ prefersNavigationBar: true }`, which is surprising.

Is there another possible name which would invert the default? An enum is also an option which gets around the problem.

foolip:
Would resolve https://github.com/whatwg/fullscreen/issues/123. @upsuper @winstonchen @jernoble, do you have thoughts about the naming/semantics here?

Given a variety of default behaviors currently, does the default have to be "auto" or something which means neither "will show navigator bar" nor "will not show navigation bar"?

jernoble:
Given the choice between default == false and default == true, I’m biased towards default == false, which would imply the property should be `preferNoNavigationBar`.

But this may be a false choice; another option would be to target the use case and not the behavior. So, given the discussion in #123, this could be `prefersMinimalUI: true` or `presentationStyle: “video”` or something along those lines. Samsung could interpret that as “hide the nav bar”, without actually writing “navigation bar” into the spec.

dtapuska:
Ya the other thing I tossed around was immersiveMode which matches the Android UI documents. However on Android immersiveMode usually implies a hidden navigation bar and status bar. I believe we only want to control the navigation bar at this time.

dtapuska:
I've made the default false and renamed it `immersiveNavigationBar`. I think this addresses open concerns. Please consider merging.

annevk:
Hmm, above @foolip is saying there may or may not be a navigation bar and @jernoble was trying to avoid saying navigation bar altogether. `minimalUI = false` as a default strikes me as a better choice.

Or perhaps we can match `window.open()` terminology here? cc @zcorpan

upsuper:
On the one hand, navigation bar doesn't sound like a general concept should be put into the spec. Specifically, Android is probably the only platform has such concept. None of desktop platforms, nor iOS has it. Meanwhile, desktop browsers have a "location bar" include some navigation functionality, which makes it feel confusing in terminology.

On the other hand, for a more general concept like "minimalUI", I'm concerned about how implementations would interpret that. What does "minimal" means? Should desktop browsers show something like... system taskbar or window controls? It feels quite vague.

annevk:
If it's really only about enabling navigation I'd go with `navigationUI = false` I suppose.

dtapuska:
`navigationUI = false` is the inverse of the default we want. ie; we want by default to show the navigation bar. Preferring user experience over screen size.

dtapuska:
Ok as per discussion on IRC with @annevk adjusted API to be an enum with 3 values.

fullscreen.bs
```diff
@@ -209,7 +219,22 @@ if all of the following are true, and false otherwise:
  <!-- cross-process, recursive -->
 </ul>
 
-<p>The <dfn method for=Element><code>requestFullscreen()</code></dfn> method, when invoked, must run
+<p>The {{FullscreenOptions}} dictionary can be used to help customize the User Agent UI:
```
annevk: I suspect @foolip will comment on this as well, but this should be defined as part of the method definition.

fullscreen.bs
```diff
@@ -180,8 +190,8 @@ partial interface DocumentOrShadowRoot {
      interface, which is implemented by Document and HTMLElement, not Element. -->
 
 <dl class=domintro>
- <dt><code><var>promise</var> = <var>element</var> . {{Element/requestFullscreen()}}</code>
- <dd><p>Displays <var>element</var> fullscreen and resolves <var>promise</var> when done.
+ <dt><code><var>promise</var> = <var>element</var> . {{Element/requestFullscreen(<var>options</var>)}}</code>
+ <dd><p>Displays <var>element</var> fullscreen with <var>options</var> and resolves <var>promise</var> when done.
```
annevk: This should describe the options members in a way that conveys the meaning to web developers.

fullscreen.bs
```diff
@@ -155,8 +155,18 @@ steps:
...
 <pre class=idl>
+enum <dfn>FullscreenNavigationUI</dfn> {
+  "auto",
+  "show",
+  "hide"
+};
+
+dictionary <dfn>FullscreenOptions</dfn> {
+  FullscreenNavigationUI navigationUI = "auto";
+};
```
annevk: I'd prefer placing these at the end of the IDL block. It's more common to start with the concrete things than the primitives, I think.

foolip: I believe there's quite a mix on that front. I checked 5 cases at "random" before getting bored. Dictionaries first:
* https://drafts.csswg.org/cssom-view/#extensions-to-the-window-interface
* https://encoding.spec.whatwg.org/#interface-textdecoder
Dictionaries last:
* https://dom.spec.whatwg.org/#interface-eventtarget
* https://wicg.github.io/background-fetch/#background-fetch-manager
* https://w3c.github.io/sensors/#the-sensor-interface

I'll go with Dave's suggested order here since it reads fine to me.

foolip:
Basic test: https://github.com/web-platform-tests/wpt/pull/12828

fullscreen.bs
```diff
@@ -254,11 +271,17 @@ these steps:
...
- <li><p>If <var>error</var> is false: Resize <var>pendingDoc</var>'s
- <a>top-level browsing context</a>'s <a>active document</a>'s viewport's dimensions to match the
- dimensions of the screen of the output device. Optionally display a message how the end user can
- revert this.
- <!-- cross-process -->
+ <li>
+  <p>If <var>error</var> is false: Resize <var>pendingDoc</var>'s
+  <a>top-level browsing context</a>'s <a>active document</a>'s viewport's dimensions to match the
+  dimensions of the available screen of the output device, optionally taking into account
+  <var>options</var>'s <code>navigationUI</code> member. Optionally display a message how the end
```
annevk: You should describe here what the user agent might use the values for. Currently you'd have to infer it from all kinds of non-normative text you could ordinarily ignore.

foolip:
@annevk @dtapuska spelled out what the UA should do if it *does* do the optional behavior in a table, PTAL.

foolip:
This is now blocked only on the Travis failure. I'll fix in https://github.com/whatwg/fullscreen/pull/137 and then rebase this.

Issue #152, opened 2019-05-13
Consume user gestures in requestFullscreen


foolip:
There is a Chromium (non-public) spoofing bug involving `requestFullscreen` that makes it seem like a good idea to consume user gestures in `requestFullscreen` as a mitigation. There may be web compat issues with such a change so it will be tested before a spec change is proposed in earnest.

This is a tracking bug for the tentative tests that may go with this change, so that they don't remain tentative forever.

@SummerLW @mustaqahmed

upsuper:
What do you mean by consume user gesture?

annevk:
cc @EdgarChen

mustaqahmed:
Consuming user activation (or user gesture) means after `requestFullscreen()` is successful, the user activation is "gone" so another `requestFullscreen()` call will fail unless there is another user activation in-between.  Essentially we are proposing one call per user interaction.  Currently any number of calls can be made until the user activation expires, which seems counter-intuitive (and it is one reason the UI spoofing bug was possible).

A bit of a background: here are [three different ways](https://mustaqahmed.github.io/user-activation-v2/#classifying-user-activation-gated-apis) APIs typically depend on user activation.

upsuper:
Sounds like a reasonable idea.

Another question would be whether other operations which require user activation / user gesture consumes that as well? i.e. if I trigger something else in user activation, and then `requestFullscreen()` should that still work?

SummerLW:
Other operation like pop up windows consume user activation. For the same
user activation, if the pop up window opens first and it consume the user
activation, the requestFullScreen() will not work. This is defined in the
current user activation v2 spec.

foolip:
FWIW, the main compat risk I see with this is that people might call `video.requestFullscreen()` + `video.play()` in the same event handler, and that consuming the gesture will mean the video won't play. Combination with other APIs that consume user activation are also possible of course, but we'll have to see which come up in practice.

@mustaqahmed @SummerLW do you anticipate this type of problem, and is there a fix we can make if it break sites?

mustaqahmed:
Yes, we expect problems like this.  One possible way to fix this is to reverse the order of the two calls, so replacing:
  `video.requestFullscreen()` then `video.play()`
with
  `video.play()` then `video.requestFullscreen()`
[...]

PR #153, opened 2019-05-28
Consume user activation in `element.requestFullscreen()`

foolip:
Depends on https://github.com/whatwg/html/pull/3851.

Fixes https://github.com/whatwg/fullscreen/issues/152.

   * Chromium: https://github.com/whatwg/fullscreen/pull/153#issuecomment-496521939
   * Gecko: https://github.com/whatwg/fullscreen/pull/153#issuecomment-1298746998
   * https://github.com/web-platform-tests/wpt/pull/36969
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1804635
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=247920
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: https://github.com/mdn/content/issues/22303

***
<a href="https://whatpr.org/fullscreen/153.html" title="Last updated on May 3, 2023, 7:23 AM UTC (e1ab822)">Preview</a> | <a href="https://whatpr.org/fullscreen/153/a69e295...e1ab822.html" title="Last updated on May 3, 2023, 7:23 AM UTC (e1ab822)">Diff</a>

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -266,6 +266,9 @@ are:
    algorithm is <a>triggered by a user generated orientation change</a>.
   </ul>
 
+ <li><p>If <var>error</var> is false, then <a>consume user activation</a> given
+ <var>pendingDoc</var>'s <a>relevant global object</a>.
+
  <li><p>Return <var>promise</var>, and run the remaining steps <a>in parallel</a>.
 
  <li>


mustaqahmed:
Thanks, this looks good from user activation perspective.

SummerLW:
https://github.com/web-platform-tests/wpt/pull/16758 has been merged into wpt repository.
Three new tests about this behavior are now in https://w3c-test.org/fullscreen/api/
element-request-fullscreen-twice-manual.tentative.html
element-request-fullscreen-two-elements-manual.tentative.html
element-request-fullscreen-two-iframes-manual.tentative.html

fullscreen.bs
```diff
@@ -270,6 +270,9 @@ when invoked, must run these steps:
    <li><p>This algorithm is <a>allowed to request fullscreen</a>.
   </ul>
 
+ <li><p>If <var>error</var> is false, then <a>consume user activation</a> for
```
mustaqahmed: FYI: This should now point to the newly added definition of [consumption](https://html.spec.whatwg.org/multipage/interaction.html#consume-user-activation).  Not sure why the PR preview doesn't show a link.

mustaqahmed: Looks like the preview is correctly pointing to the HTML spec link now.

mustaqahmed (commented):
I wanted to approve this PR, but apparently I don't have permission to do that!  @domenic?

fullscreen.bs
```diff
@@ -270,6 +270,9 @@ when invoked, must run these steps:
...
+ <var>pendingDoc</var>'s <a>browsing context</a>.
```
domenic, line 274: This should be pendingDoc's relevant global object, not its browsing context.

domenic (changes requested):

domenic:
@foolip, can we help move this along? I'm happy to rebase it and also take care of any changes after #208.

This would probably solve the issue @marcoscaceres raises in https://github.com/web-platform-tests/wpt/issues/36727 of giving a cross-browser way of consuming user activation.

marcoscaceres:
I guess it depends on what features/API calls follow a call to `.requestFullScreen()`... I see there is mention of Pointerlock in the bug and Screen Orientation Lock is another.  

I'm simultaneously hoping we can prevent the Screen Orientation API (which depends on FS API) from arbitrarily changing the screen orientation [without a user gesture](https://github.com/w3c/screen-orientation/pull/218). 

If you have an Android device handy, try this out in Chrome: 
https://marcoscaceres.github.io/playground/annoying.html 

(if you can't try it out, it loops rapidly changing the screen orientation making it difficult of users to access the browsers UI - you need to either turn off the screen or quickly swipe from the bottom to stop it). 

I personally don't think the above behavior should be acceptable for any API to do on the Web. So, I was hoping we wouldn't consume the user activation in FS API, but rather have the Screen Orientation API consume the user activation. 

Alternatively, we could have it consume the user activation if #186 was in place, all in one go. The advantage being that it wouldn't breaking any sites that are depending on FS API to not consuming the user activation.

For pointerlock, maybe a solution like #186 could also work. You request them all at once, and exiting fulls screen releases all the locks.

EdgarChen:
Currently, Gecko also consume the user activation as we had received a bug of trapping the user in fullscreen mode, which invites DOS and spoofing possibilities. And for pointer lock, we allow it without a vaild user activation if document is in fullscreen mode.

marcoscaceres:
Discussing with my WebKit colleagues... but if both Gecko and Chromium consume it, we may follow suit. Let me get back to you. 

> And for pointer lock, we allow it without a vaild user activation if document is in fullscreen mode.

That is contrary to what is in the Pointer Lock spec, right? Why don't we just link the actions together via a mechanism like in #186? 

(also, we should really rewrite the pointer lock spec to integrate properly with the Full Screen API)

mustaqahmed:
> Currently, Gecko also consume the user activation as we had received a bug of trapping the user in fullscreen mode, which invites DOS and spoofing possibilities.

Chrome originally proposed this consumption behavior to prevent a spoof, see https://crbug.com/851302.

>> And for pointer lock, we allow it without a vaild user activation if document is in fullscreen mode.
>
> That is contrary to what is in the Pointer Lock spec, right? Why don't we just link the actions together via a mechanism like in https://github.com/whatwg/fullscreen/issues/186?

IIRC Chrome added a similar hack to support some compat argument, can't recall any details!  Yes, #186 would be nice.

EdgarChen:
https://github.com/whatwg/fullscreen/issues/186 sounds reasonable to me.

marcoscaceres:
@EdgarChen, if we go with something like #207 (new dictionary member), would you be willing to remove the pointerLock workaround in Firefox? That way, we can just have a consistent model across three specs. 

Similar question for Chrome folks. I'm willing to spec that up if everyone is supportive.

fullscreen.bs
```diff
@@ -266,6 +266,9 @@ are:
    algorithm is <a>triggered by a user generated orientation change</a>.
   </ul>
 
+ <li><p>If <var>error</var> is false, then <a>consume user activation</a> given
+ <var>pendingDoc</var>'s <a>node navigable</a>'s <a>active window</a>.
```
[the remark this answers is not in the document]
domenic: Use `<a for=navigable>activate window</a>`

EdgarChen:
> @EdgarChen, if we go with something like #207 (new dictionary member), would you be willing to remove the pointerLock workaround in Firefox?

To clarify this a bit more: This is not a workaround added for consuming user activation, we behave the same all the way back to Firefox 22 at least (before the current user activation model in spec is proposed). I am worried that removing this would cause compatibility problem. :(

foolip:
@marcoscaceres there are 3 tests in wpt that are tentative due to #152:

https://github.com/web-platform-tests/wpt/blob/master/fullscreen/api/element-request-fullscreen-twice-manual.tentative.html
https://github.com/web-platform-tests/wpt/blob/master/fullscreen/api/element-request-fullscreen-two-elements-manual.tentative.html
https://github.com/web-platform-tests/wpt/blob/master/fullscreen/api/element-request-fullscreen-two-iframes-manual.tentative.html

But I think with consuming, this can be reduced to just the first test, ensuring the second call is rejected.

fullscreen.bs
```diff
@@ -266,6 +266,9 @@ are:
    algorithm is <a>triggered by a user generated orientation change</a>.
   </ul>
 
+ <li><p>If <var>error</var> is false, then <a>consume user activation</a> given
+ <var>pendingDoc</var>'s <a>node navigable</a>'s <a for=navigable>active window</a>.
```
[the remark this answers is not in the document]
domenic: Yeah, looks like that's not properly exported. I'll get on it.

marcoscaceres:
Filed WebKit bug:
https://bugs.webkit.org/show_bug.cgi?id=247920

Thanks for the details about the tests, @foolip. If those can get updated and changed to non-tentative, that would be great.

upsuper:
I have a feeling that there is nothing wrong to allow `requestFullscreen` without user activation when the document is already in fullscreen, or is entering fullscreen (by a previous call). Though to be honest I'm not sure whether it is useful to allow that either.

foolip:
@upsuper I think you're right about that, since the page script already has full control over what's being presented and could replace all the contents of the current fullscreen element. However, that'd be a change from the currently implemented behavior, and I'd prefer to handle that separately from this. Can you file an issue if you think we should change that?

foolip:
@marcoscaceres test PR is up and the whole PR template is filled in now.

marcoscaceres:
Thanks @foolip! Everything looks great.

fullscreen.bs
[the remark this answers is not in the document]
domenic: Actually you should just use `<var>pendingDoc</var>'s <a>relevant global object</a>`! No need to go all the way to navigable and back.

marcoscaceres, line 270:
```suggestion
 <li><p>If <var>error</var> is false, then <a>consume user activation</a> given
 <var>pendingDoc</var>'s <a>relevant global object</a>.
```

marcoscaceres:
This is now also implemented in WebKit. 

@foolip, anything left to get this merged?

fullscreen.bs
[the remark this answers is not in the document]
foolip: Ah, thanks!

foolip: Applied, thanks!

foolip:
@marcoscaceres sorry I totally dropped the ball here. I've updated the PR and all I need now is review from someone with write access. Paging @annevk, as @domenic seems to be OOO today.

annevk (approved):
This seems fine. The step immediately preceding this could do with some love at some point. The booleans are still hurting me.

foolip:
Thanks @annevk! Let me try inverting the preceding steps and see if you like that.

PR #177, opened 2020-06-29
Meta: update repository files

foolip:
See https://github.com/whatwg/spec-factory for details.


foolip:
The build failure is being fixed in https://github.com/whatwg/fullscreen/pull/178.

PR #181, opened 2020-07-20
Meta: adapt to the in-progress Permissions Policy renaming

foolip:
The spec URL and its entry in Specref has not changed, so there it
remains "feature" instead of "permissions".

***
<a href="https://whatpr.org/fullscreen/181.html" title="Last updated on Jul 20, 2020, 10:28 AM UTC (afcb0d2)">Preview</a> | <a href="https://whatpr.org/fullscreen/181/1f4ed06...afcb0d2.html" title="Last updated on Jul 20, 2020, 10:28 AM UTC (afcb0d2)">Diff</a>

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -695,22 +695,23 @@ iframe:fullscreen {
 
 
 
-<h2 id=feature-policy-integration>Feature Policy Integration</h2>
+<h2 id=permissions-policy-integration oldids=feature-policy-integration>Permissions Policy
+Integration</h2>
 
 <p>This specification defines a <a>policy-controlled feature</a> identified by the string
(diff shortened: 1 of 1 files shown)


foolip:
@clelland are you tracking renaming the URL and the Specref entry also? It would be great if that happens fast so other specs only need to deal with this rename once.

clelland:
I've just had the repository moved, and should be updating SpecRef today, hopefully.

foolip:
@clelland excellent, thank you!

clelland:
SpecRef changes are merged now; the old name is now an alias to permissions-policy-1, which should reduce any breakage with existing references.

PR #194, opened 2021-07-13
Editorial: replace "context object" with "this"

domenic:
Plus other minor modernizations.

***
<a href="https://whatpr.org/fullscreen/194.html" title="Last updated on Jul 13, 2021, 8:30 PM UTC (e8b152a)">Preview</a> | <a href="https://whatpr.org/fullscreen/194/1100c9f...e8b152a.html" title="Last updated on Jul 13, 2021, 8:30 PM UTC (e8b152a)">Diff</a>

PR #195, opened 2021-07-13
Meta: add/update some definitions

domenic:
* Adds appropriate `<dfn>`s and references for navigationUI and its values
* Adds appropriate `<dfn>`s and references for fullscreenchange and fullscreenerror
* Fixes the for="" value for onfullscreenchange and onfullscreenerror to include both Document and Element, so that now the IDL block for Element links to the correct place.

(Depends on #194.)

***
<a href="https://whatpr.org/fullscreen/195.html" title="Last updated on Jul 19, 2021, 4:28 PM UTC (39ef865)">Preview</a> | <a href="https://whatpr.org/fullscreen/195/3d0e592...39ef865.html" title="Last updated on Jul 19, 2021, 4:28 PM UTC (39ef865)">Diff</a>

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -202,11 +202,12 @@ partial interface mixin DocumentOrShadowRoot {
  <dd>
   Displays <var>element</var> fullscreen and resolves <var>promise</var> when done.
 
-  When supplied, <var>options</var>'s <code>navigationUI</code> member indicates whether showing
-  navigation UI while in fullscreen is preferred or not. If set to "<code>show</code>", navigation
-  simplicity is preferred over screen space, and if set to "<code>hide</code>", more screen space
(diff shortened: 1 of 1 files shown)


fullscreen.bs
```diff
@@ -285,15 +286,15 @@ are:
     </thead>
     <tbody>
       <tr>
-        <td>"<code>hide</code>"</td>
+        <td>"<dfn enum-value for="FullscreenNavigationUI"><code>hide</code></dfn>"</td>
```
annevk, line 289: I think `{{FullscreenNavigationUI/hide}}` also works, no?

annevk (approved):

fullscreen.bs
domenic: The intention is to define it here; that would leave the definition in the IDL and reference it.

PR #209, opened 2022-11-10
Rely solely on transient activation for orientation events

marcoscaceres:
Please see:
https://github.com/whatwg/html/issues/8490

   * Chrome implements this behavior (in that they special case screen orientation change events to call `.requestFullscreen()`).
   * …
- [ ] [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at:
   * …
- [ ] [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chromium: …
   * Gecko: …
   * WebKit: …
- [ ] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: …

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -18,12 +18,6 @@ spec:infra
     type:dfn; text:string
 </pre>
 
-<pre class=anchors>
-urlPrefix: https://w3c.github.io/screen-orientation/#dfn-
-    type: dfn
-        text: triggered by a user generated orientation change
-</pre>
-
 <pre class=biblio>
 {
     "CSS": {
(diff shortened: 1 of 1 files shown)


marcoscaceres:
@foolip, the "triggered by a user generated orientation change" was underspecified and only implemented in Chrome. We are trying to address this issue in an interoperable manner in https://github.com/whatwg/html/issues/8490 

However, Screen Orientation no longer defines "triggered by a user generated orientation change" so that  doesn't make sense anymore in the FS spec.

marcoscaceres:
To be clear, there is interest to align WebKit with Chrome's behavior... but we need to specify this properly in HTML and then hook full screen API properly once that's done (independently of this PR).

Issue #201, opened 2022-08-22
Change fullscreen CSS from `top`/`bottom`/`left`/`right` to `inset`


mfreed7:
The fullscreen spec says this:

```css
*|*:not(:root):fullscreen {
  top:0 !important; right:0 !important; bottom:0 !important; left:0 !important;
}
::backdrop {
  top:0; right:0; bottom:0; left:0;
}
```

For brevity and (small) performance improvement, would it be better to do this?

```css
*|*:not(:root):fullscreen {
  inset:0 !important;
}
::backdrop {
  inset:0;
}
```

On at least Chromium, `inset:0` can be faster than individual `top`/`left`/`bottom`/`right` properties. And the notation is easier to understand. This also seems like a web-compatible change to make.

@dbaron @bfgeek

foolip:
> This also seems like a web-compatible change to make.

Is this because `top`, `right`, `bottom` and `left` if specified still override `inset`? Or what kinds of breakage are possible here?

@annevk do you know who could look into this proposal for Apple?

For Gecko it looks like @EdgarChen and @marti4d have some recent fullscreen-related changes. Do either of you know who should weigh in from Mozilla's side?

annevk:
This works for us.

I couldn't work out a way in which this would be observable, but maybe there is?

mfreed7:
> > This also seems like a web-compatible change to make.
> 
> Is this because `top`, `right`, `bottom` and `left` if specified still override `inset`? Or what kinds of breakage are possible here?

Right - I think the behavior should be identical between `inset:0` and `top:0;left:0;...`, and as @annevk said, likely not observable. (It will be "observable" in devtools.)

EdgarChen:
cc @emilio.

emilio:
Not opposed to this, and it shouldn't be observable, but I'm curious about the performance argument, is the claim that's faster to parse? I expect the performance difference to be close to none in practice, but worth the simplification I suppose.

mfreed7:
> Not opposed to this, and it shouldn't be observable, but I'm curious about the performance argument, is the claim that's faster to parse? I expect the performance difference to be close to none in practice, but worth the simplification I suppose.

@bfgeek can add more color, but it sounds like `inset:0` is faster (again, very slightly!) for the layout engine. And of course fewer tokens for the parser, but that's also very small.

Again, since it's not observable and the effect is small, I debated not even opening this issue. But I did. 😄

PR #210, opened 2022-11-24
Editorial: use inset property in the UA style sheet

foolip:
Fixes https://github.com/whatwg/fullscreen/issues/201.

   * Chromium (proposed by @mfreed7)
   * Gecko (@emilio [comment](https://github.com/whatwg/fullscreen/issues/201#issuecomment-1230813781))
   * WebKit (@annevk [comment](https://github.com/whatwg/fullscreen/issues/201#issuecomment-1230504035))
   * This isn't observable since the computed style is the same both before and after, so no test updates.
   * Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=1393369
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1802334
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=248310
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: N/A

***
<a href="https://whatpr.org/fullscreen/210.html" title="Last updated on Nov 24, 2022, 4:26 PM UTC (1bf58af)">Preview</a> | <a href="https://whatpr.org/fullscreen/210/4a1f34a...1bf58af.html" title="Last updated on Nov 24, 2022, 4:26 PM UTC (1bf58af)">Diff</a>

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -676,7 +676,7 @@ properties apply to this pseudo-element either.
 
 *|*:not(:root):fullscreen {
   position:fixed !important;
-  top:0 !important; right:0 !important; bottom:0 !important; left:0 !important;
+  inset:0 !important;
   margin:0 !important;
   box-sizing:border-box !important;
   min-width:0 !important;
@@ -698,7 +698,7 @@ iframe:fullscreen {
 
 ::backdrop {
   position:fixed;
-  top:0; right:0; bottom:0; left:0;
+  inset:0;
 }
 
 *|*:not(:root):fullscreen::backdrop {


annevk (approved):
<s>When landing this, please ensure it's prefixed with "Editorial:" to make it clear this is not intended to have any normative impact.</s>

foolip:
So I think this could be tested by checking the computed style? Right now I think `inset` will be 'auto', but it should be '0px'.

emilio:
`inset` should be `0px` now, that's how the computed style for shorthands works.

annevk:
Oh, so then the issue discussion wasn't correct. @nt1m does that change your point of view?

emilio:
I'm confused, what part of the discussion is not correct? It shouldn't be observable because `inset` being `0px` should be the behavior of both `inset` and `top` / `right` / `bottom` / `left`.

annevk:
I'm also confused. Can this be tested or not? @Emilio I think you are saying it cannot be tested as `inset` should compute to `0px` before and after this change. If that's the case this can indeed go ahead.

emilio:
Yes, that's what I'm saying, sorry if I was unclear.

foolip:
Sorry, I made some assumptions without testing in actual fullscreen. I can confirm the fullscreen element's computed `inset` is "0px" in Chrome before this change, where I assumed it'd be "auto" like I saw on a test page not in fullscreen.

If that's the case then I don't think this could be observed or tested, and I'll label it editorial.

foolip:
I've filed bugs to follow this change and will merge now.

PR #204, opened 2022-10-18
Hide popovers when making an element fullscreen

josepharhar:
This is part of the new popover attribute: https://github.com/whatwg/html/pull/8221

- [ ] At least two implementers are interested (and none opposed):
   * Chrome
   * …
- [x] [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at:
   * https://wpt.fyi/results/html/semantics/popovers/popover-top-layer-combinations.html
   * https://wpt.fyi/results/html/semantics/popovers/popover-top-layer-interactions.html
- [ ] [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chromium: Implemented
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1827225
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=255232
- [ ] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: …

***
<a href="https://whatpr.org/fullscreen/204.html" title="Last updated on May 1, 2023, 2:56 PM UTC (4d375fe)">Preview</a> | <a href="https://whatpr.org/fullscreen/204/a69e295...4d375fe.html" title="Last updated on May 1, 2023, 2:56 PM UTC (4d375fe)">Diff</a>

Issue #140, opened 2018-09-20
Re-introduce the hierarchy restrictions


upsuper:
The hierarchy restrictions were removed in 766dc872bcf26c012549f7bdfde21c46bebba40c and 9592913bbeca3a9d029274f310eb27ad8003622c because it seemed to be unnecessary after we moved to the top layer model.

However, it forms an unfortunate combination with the change to make `fullscreenchange` event dispatched to element, that if we have multiple leaf fullscreen element in the stack, some of the element wouldn't get the `fullscreenchange` event. I raised in [#89 (comment)](https://github.com/whatwg/fullscreen/issues/89#issuecomment-302381330), and @foolip said he's fine adding back these restrictions in [#89 (comment)](https://github.com/whatwg/fullscreen/issues/89#issuecomment-302389588) and opened #91 to do so.

Recently, #91 was abandoned probably because of lack of activity and people don't recall the reasoning to do so. So I'm opening this issue to track that.

Blink currently doesn't have the hierarchy restrictions, while Gecko still has. And if we decide to re-introduce it, Blink may want to change before it's too late.

cc @dtapuska

upsuper:
These manual tests are checking against hierarchy restrictions:
* fullscreen/api/element-ready-check-fullscreen-element-sibling-manual.html
* fullscreen/api/element-request-fullscreen-non-top-manual.html
* fullscreen/api/element-request-fullscreen-two-elements-manual.html

dtapuska:
Adding back the hierachy restrictions for this event targetting is unforunate. We are forcing users into a model of how to design their application that the content must be present in the child of the current fullscreen element.Think of a series of divs that are all siblings and you click 'next'. Which would change the fullscreen element to the next element. In order to replicate that you'd need to do a bunch of tree modifications. Also it prevents you from re-entering a fullscreen element already on the stack. 

@domenic  Adding Domenic in case he has any ideas here.

upsuper:
(For my own note: there was [a Gecko bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1195213) we can reuse if we end up not reverting the change.)

upsuper:
> Think of a series of divs that are all siblings and you click 'next'. Which would change the fullscreen element to the next element.

That is not a usecase that the current Fullscreen API is supposed to support. You would end up having lots of elements in the fullscreen layer, and you cannot exit fullscreen in one statement (unless we add #70).

[...]

PR #223, opened 2023-04-13
Remove top layer definitions, now that CSS Position 4 contains them.

tabatkins:
Also reword algos as appropriate.

As far as Fullscreen is concerned, this is *almost* a no-op change, as things are merely moving. The only normative change that's relevant here is in the "fullscreen an element" algo, as it's no longer valid to add something to the top layer when it's already there (see next para for details). At the moment I've changed it to simply check if it's already in the top layer and only add if it's not, but this can be changed as necessary. (There are some other changes to the top-layer handling as it exists in Position 4, but they're in response to CSSWG decisions and are being discussed over there.)

The issue with adding something to the top layer that's already there is that, now that we have multiple things interacting with the top layer, they can have complex interactions of their own, and arbitrarily rearranging an element put in the top layer by some other process isn't necessarily safe. For example, a popover can be part of a popover stack, with each attached to the element below. Fullscreening the bottom popover *must not* just move that bottom popover to the end of the top layer and leave the rest of the popovers behind; it has to do *something* else, probably closing the popover (and thus closing its associated stack).

I'm not 100% sure how we want to handle the interaction of these separate features. I *suspect* it'll have to be something like each defining its own "hey you're being removed from the top layer" algo, which the others can invoke before they add the element to the top layer for their own purposes. But this'll need more thought; I went with what I felt was the simplest model for right now, where if you fullscreen something already in the top layer for other reasons, it just gets its flag and that's that.

(This is technically a behavior change per spec, but doesn't appear to be a behavior change in practice, at least in Chrome - if you fullscreen an element, show a modal, then ask to fullscreen the first element again, it *does not* move on top of the dialog, it stays where it is. But opening a dialog *then* fullscreening an element does successfully put the fullscreen over the dialog. I'm having trouble getting Firefox to successfully fullscreen something so I can test this.)

/cc @mfreed7 @josepharhar @nt1m @annevk 

-------

Currently this PR won't build because I'm using a new Bikeshed metadata to ignore the MDN issue; I should have a release out supporting this new metadata either today or tomorrow. (It won't build *without* the metadata either, since it's a fatal error to be missing an MDN anchor, too.).

Potentially we could kill both by keeping the ID in the spec, but just having it explicitly say that the definition has been moved to CSS Position. Would that be better?

More generally, for all the IDs that are being moved to Position, do you want to keep them in the spec with a redirect section?

-------

   * Chrome
   * ?
   * https://github.com/web-platform-tests/wpt/pull/39828
   * Chromium: Already Chrome's behavior.
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1834542
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=257199
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: <https://github.com/mdn/browser-compat-data/issues/19367> for redirecting ::backdrop to Position; don't *think* anything else needs tweaking.

***
<a href="https://whatpr.org/fullscreen/223.html" title="Last updated on May 23, 2023, 2:53 PM UTC (2ff6c60)">Preview</a> | <a href="https://whatpr.org/fullscreen/223/afd56a3...2ff6c60.html" title="Last updated on May 23, 2023, 2:53 PM UTC (2ff6c60)">Diff</a>

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -64,13 +64,21 @@ stated otherwise it is unset.
 <p>All <a for=/>documents</a> have an associated <dfn>list of pending fullscreen events</dfn>, which
 is an <a>ordered set</a> of (<a>string</a>, <a>element</a>) <a>tuples</a>. It is initially empty.
 
-<p>To <dfn>fullscreen an <var>element</var></dfn>, set <var>element</var>'s <a>fullscreen flag</a>.
-<a>Request removal from the top layer</a> given |element|'s <a>node document</a> and |element|, then
-<a>add to the top layer</a> given |element|'s <a>node document</a> and |element|.
+<p>To <dfn>fullscreen an <var>element</var></dfn>:
+
+<ol>
+ <li><p>Run <a>hide all popovers</a> given <var>element</var>'s <a>node document</a>.
+
+ <li><p>Set <var>element</var>'s <a>fullscreen flag</a>.
+
+ <li><p><a>Request removal from the top layer</a> given <var>element</var>.
+
+ <li><a>Add to the top layer</a> given <var>element</var>.
+</ol>
 
 <p>To <dfn>unfullscreen an <var>element</var></dfn>, unset <var>element</var>'s
 <a>fullscreen flag</a> and <a>iframe fullscreen flag</a> (if any), and
-<a>remove from the top layer immediately</a> given |element|'s <a>node document</a> and |element|.
+<a>remove from the top layer immediately</a> given <var>element</var>.
 
 <p>To <dfn>unfullscreen a <var>document</var></dfn>,
 <a lt="unfullscreen an element">unfullscreen</a> all <a>elements</a>, within <var>document</var>'s
@@ -114,7 +122,7 @@ these steps:
 
    <li>
     <p>If <var>document</var>'s <a>top layer</a> <a for=set>contains</a> <var>node</var>,
-    <a>remove from the top layer immediately</a> given <var>document</var> and <var>node</var>.
+    <a>remove from the top layer immediately</a> given <var>node</var>.
 
     <p class=note>Other specifications can add and remove elements from <a>top layer</a>, so
     <var>node</var> might not be <var>document</var>'s <a>fullscreen element</a>. For example,
@@ -234,6 +242,10 @@ if all of the following are true, and false otherwise:
  <li><p><var>element</var>'s <a>node document</a> is <a>allowed to use</a> the "<code><a
  data-lt="fullscreen-feature">fullscreen</a></code>" feature.
  <!-- cross-process, recursive -->
+
+ <li><p><var>element</var> <a for=Element>namespace</a> is not the <a>HTML namespace</a> or
+ <var>element</var>'s <a>popover visibility state</a> is <a for="popover visibility
+ state">hidden</a>.
 </ul>
 
 <div algorithm="requestFullscreen(options)">
@@ -269,6 +281,9 @@ are:
    <li><p><a>This</a>'s <a>relevant global object</a> has <a>transient activation</a> or the
    algorithm is <a>triggered by a user generated orientation change</a>.
   </ul>
+
+ <li><p>If <var>error</var> is false, then <a>consume user activation</a> given
+ <var>pendingDoc</var>'s <a>relevant global object</a>.
 
  <li><p>Return <var>promise</var>, and run the remaining steps <a>in parallel</a>.
 
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -72,7 +72,7 @@ is an <a>ordered set</a> of (<a>string</a>, <a>element</a>) <a>tuples</a>. It is
 
 <p>To <dfn>unfullscreen a <var>document</var></dfn>,
 <a lt="unfullscreen an element">unfullscreen</a> all <a>elements</a>, within <var>document</var>'s
-[=Document/top layer=], whose <a>fullscreen flag</a> is set.
+<a for=Document>top layer</a>, whose <a>fullscreen flag</a> is set.
 
 <hr>
 
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -660,6 +660,7 @@ allow="fullscreen *"&gt;</code>, as described in
 </div>
 
 
+
 <h2 id=security-and-privacy-considerations>Security and Privacy Considerations</h2>
 
 <p>User agents should ensure, e.g. by means of an overlay, that the end user is aware something is
@@ -675,6 +676,7 @@ allowed via permissions policy, either through the <{iframe/allowfullscreen}> at
 delivered with the <a>document</a> through which it is nested.
 
 <p>This prevents e.g. content from third parties to go fullscreen without explicit permission.
+
 
 
 <h2 id=old-links class=no-num oldids="new-stacking-layer, top-layer, top-layer-add, ::backdrop-pseudo-element, css-pe-backdrop">Previously-hosted definitions</h2>
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -677,8 +677,7 @@ delivered with the <a>document</a> through which it is nested.
 <p>This prevents e.g. content from third parties to go fullscreen without explicit permission.
 
 
-<h2 id=old-links class=no-num oldids="new-stacking-layer, top-layer, top-layer-add, ::backdrop-pseudo-element, css-pe-backdrop">
-Appendix A: Previously-Hosted Definitions</h2>
+<h2 id=old-links class=no-num oldids="new-stacking-layer, top-layer, top-layer-add, ::backdrop-pseudo-element, css-pe-backdrop">Previously-hosted definitions</h2>
 
 This specification previously hosted the definitions of <a selector>::backdrop</a>
 and the concept of the document's <a>top layer</a>.
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -16,7 +16,9 @@ spec:dom
 spec:infra
     type:dfn; for:set; text:for each
     type:dfn; text:string
-spec:css-position-4; type:selector; text: ::backdrop
+spec:css-position-4
+    type:selector; text: ::backdrop
+    type:dfn; text:top layer
 </pre>
 
 <pre class=anchors>
@@ -57,30 +59,22 @@ stated otherwise it is unset.
 
 <p>All <a for=/>documents</a> have an associated <dfn export>fullscreen element</dfn>. The
 <a>fullscreen element</a> is the topmost <a>element</a> in the <a for=/>document</a>'s
-<a for=Document>top layer</a> whose <a>fullscreen flag</a> is set, if any, and null otherwise.
+<a>top layer</a> whose <a>fullscreen flag</a> is set, if any, and null otherwise.
 
 <p>All <a for=/>documents</a> have an associated <dfn>list of pending fullscreen events</dfn>, which
 is an <a>ordered set</a> of (<a>string</a>, <a>element</a>) <a>tuples</a>. It is initially empty.
 
-<p>To <dfn>fullscreen an <var>element</var></dfn>:
-
-<ol>
- <li><p>Run <a>hide all popovers</a> given <var>element</var>'s <a>node document</a>.
-
- <li><p>Set <var>element</var>'s <a>fullscreen flag</a>.
-
- <li><p><a>Request removal from the top layer</a> given <var>element</var>.
-
- <li><a>Add to the top layer</a> given <var>element</var>.
-</ol>
+<p>To <dfn>fullscreen an <var>element</var></dfn>, set <var>element</var>'s <a>fullscreen flag</a>.
+<a>Request removal from the top layer</a> given |element|'s <a>node document</a> and |element|, then
+<a>add to the top layer</a> given |element|'s <a>node document</a> and |element|.
 
 <p>To <dfn>unfullscreen an <var>element</var></dfn>, unset <var>element</var>'s
 <a>fullscreen flag</a> and <a>iframe fullscreen flag</a> (if any), and
-<a>remove from the top layer immediately</a> given <var>element</var>.
+<a>remove from the top layer immediately</a> given |element|'s <a>node document</a> and |element|.
 
 <p>To <dfn>unfullscreen a <var>document</var></dfn>,
 <a lt="unfullscreen an element">unfullscreen</a> all <a>elements</a>, within <var>document</var>'s
-<a for=Document>top layer</a>, whose <a>fullscreen flag</a> is set.
+<a>top layer</a>, whose <a>fullscreen flag</a> is set.
 
 <hr>
 
@@ -91,7 +85,7 @@ is an <a>ordered set</a> of (<a>string</a>, <a>element</a>) <a>tuples</a>. It is
  <li><p>If <var>document</var>'s <a>fullscreen element</a> is null, terminate these steps.
 
  <li><p><a lt="unfullscreen an element">Unfullscreen elements</a> whose <a>fullscreen flag</a> is
- set, within <var>document</var>'s <a for=Document>top layer</a>, except for <var>document</var>'s
+ set, within <var>document</var>'s <a>top layer</a>, except for <var>document</var>'s
  <a>fullscreen element</a>.
 
  <li><p><a>Exit fullscreen</a> <var>document</var>.
@@ -119,10 +113,10 @@ these steps:
    <li><p>Otherwise, <a lt="unfullscreen an element">unfullscreen <var>node</var></a>.
 
    <li>
-    <p>If <var>document</var>'s <a for=Document>top layer</a> <a for=set>contains</a> <var>node</var>,
-    <a>remove from the top layer immediately</a> given <var>node</var>.
+    <p>If <var>document</var>'s <a>top layer</a> <a for=set>contains</a> <var>node</var>,
+    <a>remove from the top layer immediately</a> given <var>document</var> and <var>node</var>.
 
-    <p class=note>Other specifications can add and remove elements from <a for=Document>top layer</a>, so
+    <p class=note>Other specifications can add and remove elements from <a>top layer</a>, so
     <var>node</var> might not be <var>document</var>'s <a>fullscreen element</a>. For example,
     <var>node</var> could be an open <{dialog}> element.
   </ol>
@@ -240,10 +234,6 @@ if all of the following are true, and false otherwise:
  <li><p><var>element</var>'s <a>node document</a> is <a>allowed to use</a> the "<code><a
  data-lt="fullscreen-feature">fullscreen</a></code>" feature.
  <!-- cross-process, recursive -->
-
- <li><p><var>element</var> <a for=Element>namespace</a> is not the <a>HTML namespace</a> or
- <var>element</var>'s <a>popover visibility state</a> is <a for="popover visibility
- state">hidden</a>.
 </ul>
 
 <div algorithm="requestFullscreen(options)">
@@ -279,9 +269,6 @@ are:
    <li><p><a>This</a>'s <a>relevant global object</a> has <a>transient activation</a> or the
    algorithm is <a>triggered by a user generated orientation change</a>.
   </ul>
-
- <li><p>If <var>error</var> is false, then <a>consume user activation</a> given
- <var>pendingDoc</var>'s <a>relevant global object</a>.
 
  <li><p>Return <var>promise</var>, and run the remaining steps <a>in parallel</a>.
 
@@ -417,9 +404,9 @@ getter steps are:
 </div>
 
 <p>A <a>document</a> is said to be a <dfn>simple fullscreen document</dfn> if there is exactly one
-<a>element</a> in its <a for=Document>top layer</a> that has its <a>fullscreen flag</a> set.
+<a>element</a> in its <a>top layer</a> that has its <a>fullscreen flag</a> set.
 
-<p class=note>A <a>document</a> with two <a>elements</a> in its <a for=Document>top layer</a> can be a
+<p class=note>A <a>document</a> with two <a>elements</a> in its <a>top layer</a> can be a
 <a>simple fullscreen document</a>. For example, in addition to the <a>fullscreen element</a> there
 could be an open <{dialog}> element.
 
@@ -453,7 +440,7 @@ could be an open <{dialog}> element.
 
  <p class=note>This is the set of documents for which the <a>fullscreen element</a> will be
  <a lt="unfullscreen an element">unfullscreened</a>, but the last document in <var>docs</var> might
- have more than one <a>element</a> in its <a for=Document>top layer</a> with the <a>fullscreen flag</a> set,
+ have more than one <a>element</a> in its <a>top layer</a> with the <a>fullscreen flag</a> set,
  in which case that document will still remain in fullscreen.
 </ol>
 </div>
@@ -679,7 +666,7 @@ delivered with the <a>document</a> through which it is nested.
 Appendix A: Previously-Hosted Definitions</h2>
 
 This specification previously hosted the definitions of <a selector>::backdrop</a>
-and the concept of the document's <a for=Document>top layer</a>.
+and the concept of the document's <a>top layer</a>.
 
 
 
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -665,7 +665,7 @@ delivered with the <a>document</a> through which it is nested.
 Appendix A: Previously-Hosted Definitions</h2>
 
 This specification previously hosted the definitions of <a selector>::backdrop</a>
-and the concept of the [=Document/top layer=].
+and the concept of the document's [=Document/top layer=].
 
 
 
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -71,7 +71,7 @@ is an <a>ordered set</a> of (<a>string</a>, <a>element</a>) <a>tuples</a>. It is
 
  <li><p>Set <var>element</var>'s <a>fullscreen flag</a>.
 
- <li><p><a>Request removal from the top layer</a> given <var>element</var>.
+ <li><p><a>Remove from the top layer immediately</a> given <var>element</var>.
 
  <li><a>Add to the top layer</a> given <var>element</var>.
 </ol>
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -69,14 +69,14 @@ is an <a>ordered set</a> of (<a>string</a>, <a>element</a>) <a>tuples</a>. It is
 
  <li><p>Set <var>element</var>'s <a>fullscreen flag</a>.
 
- <li><p><a>Request removal from the top layer</a> given |element|'s <a>node document</a> and |element|.
+ <li><p><a>Request removal from the top layer</a> given <var>element</var>.
 
- <li><a>Add to the top layer</a> given |element|'s <a>node document</a> and |element|.
+ <li><a>Add to the top layer</a> given <var>element</var>.
 </ol>
 
 <p>To <dfn>unfullscreen an <var>element</var></dfn>, unset <var>element</var>'s
 <a>fullscreen flag</a> and <a>iframe fullscreen flag</a> (if any), and
-<a>remove from the top layer immediately</a> given |element|'s <a>node document</a> and |element|.
+<a>remove from the top layer immediately</a> given <var>element</var>.
 
 <p>To <dfn>unfullscreen a <var>document</var></dfn>,
 <a lt="unfullscreen an element">unfullscreen</a> all <a>elements</a>, within <var>document</var>'s
@@ -120,7 +120,7 @@ these steps:
 
    <li>
     <p>If <var>document</var>'s <a for=Document>top layer</a> <a for=set>contains</a> <var>node</var>,
-    <a>remove from the top layer immediately</a> given <var>document</var> and <var>node</var>.
+    <a>remove from the top layer immediately</a> given <var>node</var>.
 
     <p class=note>Other specifications can add and remove elements from <a for=Document>top layer</a>, so
     <var>node</var> might not be <var>document</var>'s <a>fullscreen element</a>. For example,
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -7,7 +7,6 @@ Text Macro: LATESTRD 2023-01
 Abstract: The Fullscreen API standard defines an API for elements to display themselves fullscreen.
 Translation: ja https://triple-underscore.github.io/fullscreen-ja.html
 Markup Shorthands: css no
-Ignore MDN Failure: ::backdrop-pseudo-element
 </pre>
 
 <pre class=link-defaults>
@@ -17,6 +16,7 @@ spec:dom
 spec:infra
     type:dfn; for:set; text:for each
     type:dfn; text:string
+spec:css-position-4; type:selector; text: ::backdrop
 </pre>
 
 <pre class=anchors>
@@ -659,6 +659,13 @@ allowed via permissions policy, either through the <{iframe/allowfullscreen}> at
 delivered with the <a>document</a> through which it is nested.
 
 <p>This prevents e.g. content from third parties to go fullscreen without explicit permission.
+
+
+<h2 id=old-links class=no-num oldids="new-stacking-layer, top-layer, top-layer-add, ::backdrop-pseudo-element, css-pe-backdrop">
+Appendix A: Previously-Hosted Definitions</h2>
+
+This specification previously hosted the definitions of <a selector>::backdrop</a>
+and the concept of the [=Document/top layer=].
 
 
 
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -63,7 +63,7 @@ stated otherwise it is unset.
 is an <a>ordered set</a> of (<a>string</a>, <a>element</a>) <a>tuples</a>. It is initially empty.
 
 <p>To <dfn>fullscreen an <var>element</var></dfn>, set <var>element</var>'s <a>fullscreen flag</a>.
-If <var>element</var> is not currently in its <a>node document</a>'s [=Document/top layer=],
+If <var>element</var> is not currently [=in the top layer=] of its <a>node document</a>,
 [=add to the top layer|add it to the top layer=] of its <a>node document</a>.
 
 <p>To <dfn>unfullscreen an <var>element</var></dfn>, unset <var>element</var>'s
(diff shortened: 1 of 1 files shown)


foolip:
@tabatkins the Bikeshed build is still failing. Did you release a new Bikeshed yet?

foolip:
I think the prose here looks good and it's just a question of what the behavior should be in two cases. For the "already in top layer case", I think it's a very rare situation in practice, but there are the situations I can think of worth testing:
- Requesting fullscreen on the same element twice (should be unaffected)
- Switching between two elements in fullscreen by calling `a.requestFullscreen()` and `b.requestFullscreen()` in turns, over and over. This would have worked before this change, but not any longer.
- An open `<dialog>` or `popver` element in top layer in some position other than the top, so most likely nested popovers and requesting fullscreen on a popover that isn't currently on top.

@tabatkins what do you think about the second case here?

To also respond to your PR description:

> I'm having trouble getting Firefox to successfully fullscreen something so I can test this.

Do you have the test code available that you were trying?

> Potentially we could kill both by keeping the ID in the spec, but just having it explicitly say that the definition has been moved to CSS Position. Would that be better?

Do you have open PRs for all specs that link to the IDs that will go away? If so I think it's fine to let the cross-spec links be broken for a short time.

> More generally, for all the IDs that are being moved to Position, do you want to keep them in the spec with a redirect section?

I would say no. @annevk how have you done things for other specs?

tabatkins:
> Did you release a new Bikeshed yet?

No, I was blocked on <https://github.com/whatwg/meta/issues/271>, but I've gone ahead and unblocked myself there. The Bikeshed release is now public, 3.11.22.

> Do you have the test code available that you were trying?

I was just doing `el.requestFullScreen()` in the console on the Fullscreen spec's logo, along with a manually inserted `dialog` element.

> Do you have open PRs for all specs that link to the IDs that will go away? If so I think it's fine to let the cross-spec links be broken for a short time.

No, I need to figure out how to get this info, I think it's available from WebRef.

tabatkins:
> Requesting fullscreen on the same element twice (should be unaffected)

Yes, it's unaffected afaict.

> Switching between two elements in fullscreen by calling a.requestFullscreen() and b.requestFullscreen() in turns, over and over. This would have worked before this change, but not any longer.

Hm, we can make this work if necessary.

> An open <dialog> or popver element in top layer in some position other than the top, so most likely nested popovers and requesting fullscreen on a popover that isn't currently on top.

You already can't fullscreen a dialog, and <https://github.com/whatwg/fullscreen/pull/204> closes popovers when an element fullscreens.

fullscreen.bs
```diff
@@ -56,21 +57,22 @@ stated otherwise it is unset.
...
-and <a for="top layer">add</a> it to its <a>node document</a>'s <a>top layer</a>.
+<p>To <dfn>fullscreen an <var>element</var></dfn>, set <var>element</var>'s <a>fullscreen flag</a>.
+If <var>element</var> is not currently [=in the top layer=] of its <a>node document</a>,
+[=add to the top layer|add it to the top layer=] of its <a>node document</a>.
 
 <p>To <dfn>unfullscreen an <var>element</var></dfn>, unset <var>element</var>'s
-<a>fullscreen flag</a> and <a>iframe fullscreen flag</a> (if any), and <a for=set>remove</a> it from
-its <a>node document</a>'s <a>top layer</a>.
+<a>fullscreen flag</a> and <a>iframe fullscreen flag</a> (if any), and
+<a>request removal from the top layer</a> for it from its <a>node document</a>.
```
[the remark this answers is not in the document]
tabatkins: Yes, `:fullscreen` will immediately stop matching. I think this is good - it means if you have transitions set up for an exiting-fullscreen animation, they'll start going then. The difference between "Immediately" and "at next style update" aren't significant for this purpose; top layer needs to delay so it can react to newly-added 'overlay' transitions, but otherwise the CSS effects both happen at the same time, when you Update The Rendering.

fullscreen.bs
```diff
@@ -56,21 +57,22 @@ stated otherwise it is unset.
...
-<a>top layer</a> whose <a>fullscreen flag</a> is set, if any, and null otherwise.
+[=Document/top layer=] whose <a>fullscreen flag</a> is set, if any, and null otherwise.
 
 <p>All <a for=/>documents</a> have an associated <dfn>list of pending fullscreen events</dfn>, which
 is an <a>ordered set</a> of (<a>string</a>, <a>element</a>) <a>tuples</a>. It is initially empty.
 
-<p>To <dfn>fullscreen an <var>element</var></dfn>, set <var>element</var>'s <a>fullscreen flag</a>
-and <a for="top layer">add</a> it to its <a>node document</a>'s <a>top layer</a>.
+<p>To <dfn>fullscreen an <var>element</var></dfn>, set <var>element</var>'s <a>fullscreen flag</a>.
+If <var>element</var> is not currently [=in the top layer=] of its <a>node document</a>,
```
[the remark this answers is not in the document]
tabatkins: It seems like the behavior we're going for, actually, is that an element can only be in the top layer for one reason at a time, so actually this might need to change a little more to properly reflect that. (And possibly allow it to be moved if it's currently in the top layer for fullscreen reasons, as per your comment.)

foolip:
Thanks @tabatkins, this now builds!

I've poked https://github.com/whatwg/fullscreen/pull/204 to see if we can get that landed, although it doesn't block this PR.

I think we can put interactions between fullscreen and other top layer-using features to the side, since as you point out `<dialog>` already can't go fullscreen, and popovers are new so there's no site compat risk to consider.

That leaves pure uses of the Fullscreen API with multiple elements. I think the conservative thing to do is to conserve the existing behavior where we move an element within the top layer. I'm not aware of real-world uses of it, but it's not especially far-fetched and the web is big.

I think the most straightforward approach is to call "remove an element from the top layer immediately" followed by " an element to the top layer", is that what you think we should do?

tabatkins:
Yeah, if we currently have compat with `A.rFS(); B.rFS(); A.rFS();` working and ending with A fullscreened on top, then I think we should indeed just edit that behavior in. I suspect the check should be that A is in the top layer *and* has the fullscreen flag, but isn't the doc's fullscreen element, then we remove and re-add it so it cycles to the top.

If it's already in the top layer but *doesn't* have the fullscreen flag, that means some other feature has put it in and we probably shouldn't touch it, at least not until we have the cross-feature agreement figured out.

foolip:
OK, I've tested the `A.rFS(); B.rFS(); A.rFS();` scenario with this test case:

```html
<style>
div {
    padding: 1em;
    display: grid;
    place-items: center;
}
</style>
<div id="A" style="background-color:lavender">
 <span>This is A. <button onclick="fs(event)">go/toggle fullscreen</button></span>
</div>
<div id="B" style="background-color: lemonchiffon">
 <span>This is B. <button onclick="fs(event)">go/toggle fullscreen</button></span>
</div>
<script>
const divs = Array.from(document.querySelectorAll('div'));
function fs(event) {
    let target = event.target.closest('div');
    if (target === document.fullscreenElement) {
        // Toggle between the two divs when already fullscreen.
        target = divs.find((div) => div !== document.fullscreenElement);
    }
    console.log(`Requesting fullscreen for ${target.id}`);
    target.requestFullscreen().then(() => {
        console.log(`Success for ${target.id}`);
    }, (error) => {
        console.log(`Error for ${target.id}: ${error.message}`);
    });
}
</script>
```

As it turns out, only Chrome supports toggling between fullscreen elements like this.

Firefox fails the toggle request with "Request for fullscreen was denied because requesting element is not a descendant of the current fullscreen element."

Safari also rejects the `requestFullscreen()` call, without a message, but it must be this descendant check in the source:
https://github.com/WebKit/WebKit/blob/f462028e09b021b5d2768fee9ea0ffc056565b94/Source/WebCore/dom/FullscreenManager.cpp#L224-L230

In other words, the original hierarchy restrictions are still in Gecko and WebKit, see https://github.com/whatwg/fullscreen/issues/140 for further links.

@josepharhar @mfreed7 do you have preferences on how we handle this?

annevk:
> @annevk how have you done things for other specs?

We do sometimes leave a note or small appendix for cases like this. See https://fetch.spec.whatwg.org/#websocket-protocol (and in particular the source). I think that is warranted here as well given how long Fullscreen hosted top layer.

foolip:
@tabatkins can you update this to leave a section preserving the top layer IDs pointing to their new location in https://drafts.csswg.org/css-position-4/?

tabatkins:
IDs added. I think we're just waiting on final confirmation of the decision in <https://github.com/whatwg/fullscreen/pull/223#issuecomment-1522928629>, then?

foolip:
@tabatkins that's right, let me ping folks about that.

foolip:
To help decide on the flip-flop case I checked how `<dialog>` behaves with this test:

```html
<dialog id="A" style="background-color:lavender">
    <span>This is A. <button onclick="show(event)">open/toggle dialog</button></span>
</dialog>
<dialog id="B" style="background-color: lemonchiffon">
    <span>This is B. <button onclick="show(event)">open/toggle dialog</button></span>
</dialog>
<button onclick="show(event)">open/toggle dialog</button>
<script>
const dialogs = Array.from(document.querySelectorAll('dialog'));
function show(event) {
    let dialog = dialogs.find((dialog) => !dialog.open);
    if (!dialog) {
        // Try to open the first one anyway, expect it to fail.
        dialog = dialogs[0];
    }
    console.log(`Showing dialog ${dialog.id}`);
    try {
        dialog.showModal();
        console.log(`Success for ${dialog.id}`);
    } catch (error) {
        console.log(`Error for ${dialog.id}: ${error.message}`);
    }
}
</script>
```

These are sibling `<dialog>` elements. Chrome, Firefox and Safari all open A and then B, but won't open A again because it's already open.

To get the equivalent behavior for the fullscreen API, we should reject if the element is already in top layer stack, but we _shouldn't_ add back a hierarchy restriction.

@josepharhar @mfreed7 would that work for you?

mfreed7:
> To get the equivalent behavior for the fullscreen API, we should reject if the element is already in top layer stack, but we shouldn't add back a hierarchy restriction.
> 
> @josepharhar @mfreed7 would that work for you?

If this ends up being web compatible, I'm definitely supportive. It feels right to:
 1. Try to match other top layer APIs like dialog and popover, generally. That includes:
 2. Reject/throw/no-op when a "second" top layer API call is made while an element is already in the top layer.
 3. Allow fullscreen to be flexible and not require elements to be DOM-nested in order to both use fullscreen.

foolip:
@mfreed7 sounds good! Let's hammer out the details of point 2 then, what happens for a second call to `elm.requestFullscreen()` in different scenarios.

Bottom line first: no-op success when requesting fullscreen again matches current behavior. That's what this PR already does, so no changes required. 

The investigation:

**One scenario** is calling `requestFullscreen()` twice in the same click event handler:

```js
button.onclick = () => {
  const p1 = div.requestFullscreen();
  const p2 = div.requestFullscreen();
  // Does p2 reject and why?
}
```

This is tested in [element-request-fullscreen-twice.html](https://wpt.fyi/results/fullscreen/api/element-request-fullscreen-twice.html) and Chrome, Firefox and Safari all reject the second promise. Per spec (with https://github.com/whatwg/fullscreen/pull/153 just merged) it should reject because the user gesture was consumed by the first call.

**Another scenario** is calling `requestFullscreen()` on the element that's already fullscreen, in a second click event handler:

```js
button.onclick = () => {
  const p = div.requestFullscreen();
  // Does p reject if div is document.fullscreenElement?
}
```

This is tested in [element-request-fullscreen-same-element.html](https://wpt.fyi/results/fullscreen/api/element-request-fullscreen-same-element.html). That times out in Safari, I don't know why, so I tested this manually:

```html
button.onclick = () => {
    if (!document.fullscreenElement) {
        div.requestFullscreen().then(() => {
            console.log('1st request succeeded');
        }, (error) => {
            console.log('1st request failed:', error.message);
        });
    } else {
        div.requestFullscreen().then(() => {
            console.log('2nd request succeeded');
        }, (error) => {
            console.log('2nd request failed:', error.message);
        });
    }
}
```

I've found that Chrome, Firefox and Safari all resolve the 2nd promise, in other words it's a no-op.

foolip:
For testing, a think a single new test would cover this PR. It should have two sibling divs, A and B. Request fullscreen once per click in order A-B-A and assert that all requests succeed but the last is a no-op and that `document.fullscreenElement` is still B. Bonus points for exiting fullscreen once and asserting that `document.fullscreenElement` is then A again, and then once more to make `document.fullscreenElement` null.

mfreed7:
> For testing, a think a single new test would cover this PR. It should have two sibling divs, A and B. Request fullscreen once per click in order A-B-A and assert that all requests succeed but the last is a no-op and that `document.fullscreenElement` is still B. Bonus points for exiting fullscreen once and asserting that `document.fullscreenElement` is then A again, and then once more to make `document.fullscreenElement` null.

So I wrote [a test](https://github.com/web-platform-tests/wpt/pull/39828) for this, but I'm getting a bit more concerned about the behavior. As-is, Chromium fails this test, because the current behavior is that A is ([manually](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/fullscreen/fullscreen.cc;l=203;drc=4a24affd324ec742bfb061cda42e459308e36161)) popped to the top of the top layer stack when `A.requestFullscreen()` is called the second time. Gecko and WebKit both fail before that point (at `B.requestFullscreen()`) because of the descendant check. I'm still supportive of changing this, but I'm worried about compat. Perhaps, if there's agreement with this direction (@annevk @emilio - thoughts?) then I can add a use counter for this situation and see where we stand?

The other way to go is to keep Chromium's current behavior of popping A back to the top on the second call to `requestFullscreen()`. This is definitely web compatible, and I could see an argument for this being useful for developers.

fullscreen.bs
```diff
@@ -745,6 +661,13 @@ delivered with the <a>document</a> through which it is nested.
 <p>This prevents e.g. content from third parties to go fullscreen without explicit permission.
 
 
+<h2 id=old-links class=no-num oldids="new-stacking-layer, top-layer, top-layer-add, ::backdrop-pseudo-element, css-pe-backdrop">
+Appendix A: Previously-Hosted Definitions</h2>
+
+This specification previously hosted the definitions of <a selector>::backdrop</a>
+and the concept of the [=Document/top layer=].
```
annevk, line 668:
```suggestion
and the concept of the document's [=Document/top layer=].
```

annevk (commented):
This looks okay to me. Really exciting improvement overall! Slight behavior changes are fine with me as well, but I'd like @nt1m to have the final call on those.

foolip:
@mfreed7 I think that preserving the "In other words, element is moved to the end of top layer if it is already present" behavior of the current spec has some merits. It makes for a less strange API, avoiding the situation where a request appears to succeed but it doesn't actually change anything.

For dialogs and popovers, is there any reason they shouldn't also allow this flip-flopping behavior? I don't want to go overboard with consistency, just want to check.

An important difference between fullscreen and the other features is that going fullscreen is async so the synchronous checks that `elm.showPopover()` and `dialog.showModal()` have aren't enough, there's also checks after resizing the window. So going fullscreen has the style of pre-flight check, try to do it, then deal with weird things that happened. This is all to say I think it would be OK if fullscreen is slightly different to the others in this regard.

Thoughts.

mfreed7:
> @mfreed7 I think that preserving the "In other words, element is moved to the end of top layer if it is already present" behavior of the current spec has some merits. It makes for a less strange API, avoiding the situation where a request appears to succeed but it doesn't actually change anything.

Yeah, I think I agree. And it sounds like there's support from others above as well. Sounds like we should go this direction. I'll land the WPT with that behavior (A/B/A leaves A topmost).


> For dialogs and popovers, is there any reason they shouldn't also allow this flip-flopping behavior? I don't want to go overboard with consistency, just want to check.

We could talk about dialog, I suppose. Nicely, there shouldn't be compat issues since `showModal()` throws on an already-showing dialog today.

However, I don't think we should do this flip-flopping for popovers. The popover stacking logic is complex, and I worry that we'll run into weird situations. 

In all of these cases, I really like that the "add to top layer" algorithm will not support auto-flip-flopping, and we instead have to consider what's the best behavior for each top layer consumer API.

foolip:
Thanks @mfreed7, if you advise against supporting flip-flopping popovers, then I think we should leave `<dialog>` alone as well, unless we get some developer requests for flip-flopping. And for fullscreen, supporting flip-flopping is the simplest way to make the API consistent. (To disallow it, we'd need to add code to detect the situation and reject the fullscreen request in the async part of the algorithm.)

@tabatkins can you make the change to move the element to the top of top layer if it's already present?

fullscreen.bs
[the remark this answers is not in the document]
nt1m: In practice, transitioning out of `:fullscreen` is quite difficult via CSS transitions anyway because we take out `position: fixed;`. It is also something the browser typically does a visual transition for too, which makes it hard to notice if any CSS transition is happening.

Personally, I would prefer going for "remove immediately from top layer" in this case, because exiting fullscreen is a critical operation (if someone tries to press "Esc", you don't want to allow top layer transitions) and it would be better perf wise too. Entering / exiting fullscreen is a fairly expensive operation style-wise in WebKit, especially when it involves subframes.

fullscreen.bs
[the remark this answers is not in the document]
nt1m: We already have safeguards (exceptions notably) for this so we shouldn't change behavior unless necessary. Dialogs can't be fullscreened, Popovers can't be fullscreened (with #204 merged at least). Fullscreen elements also can't be made popovers (check popover validity algorithm prevents it). They can't be made dialogs either since you can't change the tag name.

nt1m (changes requested):
I don't think we should proceed with the behavior changes, based on the comments I've added.

foolip:
@nt1m do be clear, are you talking only about the A-B-A behavior and that we should conserve the current behavior of the spec?

fullscreen.bs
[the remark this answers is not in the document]
foolip: I think here we should first check if the element is in top layer and if so remove it immediately. That will have the effect of moving it to the top of the stack.

nt1m:
> @nt1m do be clear, are you talking only about the A-B-A behavior and that we should conserve the current behavior of the spec?

I'm referring mostly to this: https://github.com/whatwg/fullscreen/pull/223/files#r1192672349

foolip:
@nt1m yes, that's exactly the A-B-A behavior I referred to, as an A-B-A sequence of requests is necessary to observe the behavior in question. I think we should preserve the existing spec behavior, which matches Chromium but not Gecko+WebKit. Is that your position too?

nt1m:
> @nt1m yes, that's exactly the A-B-A behavior I referred to, as an A-B-A sequence of requests is necessary to observe the behavior in question. I think we should preserve the existing spec behavior, which matches Chromium but not Gecko+WebKit. Is that your position too?

Yes, pretty sure that's how WebKit implements it already: https://searchfox.org/wubkat/rev/17b431999014409e4831dce71e69e0033d87ede5/Source/WebCore/dom/FullscreenManager.cpp#514-517

(If the test is failing, it's probably failing for different reasons fwiw)

mfreed7:
> (If the test is failing, it's probably failing for different reasons fwiw)

I believe it’s failing because WebKit doesn’t allow A and B to be siblings.

fullscreen.bs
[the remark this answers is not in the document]
tabatkins: Okay, I've gone ahead and applied this - when fullscreening, we first request removal *then* add it, so it'll cycle to the end if it was already there.

fullscreen.bs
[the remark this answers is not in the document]
tabatkins: Hm, that's a reasonable argument. I'm okay with removing it immediately when unfullscreening, as this is indeed somewhat of a special case.

@mfreed/@foolip does this sound okay to y'all too?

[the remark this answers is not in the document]
foolip: Yes, I think we should do that. If we want to support CSS transitions when exiting fullscreen, we can follow up on that separately. I suspect additional changes would be needed.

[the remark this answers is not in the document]
foolip:
```suggestion
<a>remove from the top layer immediately</a> given |element|'s <a>node document</a> and |element|.
```

fullscreen.bs
```diff
@@ -56,21 +57,22 @@ stated otherwise it is unset.
...
+[=Document/top layer=] whose <a>fullscreen flag</a> is set, if any, and null otherwise.
 
 <p>All <a for=/>documents</a> have an associated <dfn>list of pending fullscreen events</dfn>, which
 is an <a>ordered set</a> of (<a>string</a>, <a>element</a>) <a>tuples</a>. It is initially empty.
 
-<p>To <dfn>fullscreen an <var>element</var></dfn>, set <var>element</var>'s <a>fullscreen flag</a>
-and <a for="top layer">add</a> it to its <a>node document</a>'s <a>top layer</a>.
+<p>To <dfn>fullscreen an <var>element</var></dfn>, set <var>element</var>'s <a>fullscreen flag</a>.
+[=Request removal from the top layer=] for |element| from its [=node document=], then
+[=add to the top layer|add it to the top layer=] of its [=node document=].
```
foolip, line 67:
```suggestion
<a>Request removal from the top layer</a> given |element|'s <a>node document</a> and |element|, then
<a>add to the top layer</a> given |element|'s <a>node document</a> and |element|.
```

fullscreen.bs
```diff
@@ -56,21 +57,22 @@ stated otherwise it is unset.
 
 <p>All <a for=/>documents</a> have an associated <dfn export>fullscreen element</dfn>. The
 <a>fullscreen element</a> is the topmost <a>element</a> in the <a for=/>document</a>'s
-<a>top layer</a> whose <a>fullscreen flag</a> is set, if any, and null otherwise.
+[=Document/top layer=] whose <a>fullscreen flag</a> is set, if any, and null otherwise.
```
foolip, line 60:
```suggestion
<a for=Document>top layer</a> whose <a>fullscreen flag</a> is set, if any, and null otherwise.
```

fullscreen.bs
```diff
@@ -81,7 +83,7 @@ its <a>node document</a>'s <a>top layer</a>.
  <li><p>If <var>document</var>'s <a>fullscreen element</a> is null, terminate these steps.
 
  <li><p><a lt="unfullscreen an element">Unfullscreen elements</a> whose <a>fullscreen flag</a> is
- set, within <var>document</var>'s <a>top layer</a>, except for <var>document</var>'s
+ set, within <var>document</var>'s [=Document/top layer=], except for <var>document</var>'s
```
foolip, line 86:
```suggestion
 set, within <var>document</var>'s <a for=Document>top layer</a>, except for <var>document</var>'s
```

fullscreen.bs
```diff
@@ -109,10 +111,11 @@ these steps:
    <li><p>Otherwise, <a lt="unfullscreen an element">unfullscreen <var>node</var></a>.
 
    <li>
-    <p>If <var>document</var>'s <a>top layer</a> <a for=set>contains</a> <var>node</var>,
-    <a for=set>remove</a> <var>node</var> from <var>document</var>'s <a>top layer</a>.
+    <p>If <var>document</var>'s [=Document/top layer=] <a for=set>contains</a> <var>node</var>,
+    <a>remove from the top layer immediately</a> <var>node</var> from <var>document</var>'s
+    [=Document/top layer=].
```
foolip, line 116:
```suggestion
    <p>If <var>document</var>'s <a for=Document>top layer</a> <a for=set>contains</a> <var>node</var>,
    <a>remove from the top layer immediately</a> given <var>document</var> and <var>node</var>.
```

fullscreen.bs
```diff
@@ -109,10 +111,11 @@ these steps:
...
 
-    <p class=note>Other specifications can add and remove elements from <a>top layer</a>, so
+    <p class=note>Other specifications can add and remove elements from [=Document/top layer=], so
```
foolip, line 118:
```suggestion
    <p class=note>Other specifications can add and remove elements from <a for=Document>top layer</a>, so
```

fullscreen.bs
```diff
@@ -400,9 +403,9 @@ getter steps are:
 </div>
 
 <p>A <a>document</a> is said to be a <dfn>simple fullscreen document</dfn> if there is exactly one
-<a>element</a> in its <a>top layer</a> that has its <a>fullscreen flag</a> set.
+<a>element</a> in its [=Document/top layer=] that has its <a>fullscreen flag</a> set.
```
foolip, line 406:
```suggestion
<a>element</a> in its <a for=Document>top layer</a> that has its <a>fullscreen flag</a> set.
```

fullscreen.bs
```diff
@@ -400,9 +403,9 @@ getter steps are:
...
 
-<p class=note>A <a>document</a> with two <a>elements</a> in its <a>top layer</a> can be a
+<p class=note>A <a>document</a> with two <a>elements</a> in its [=Document/top layer=] can be a
```
foolip, line 408:
```suggestion
<p class=note>A <a>document</a> with two <a>elements</a> in its <a for=Document>top layer</a> can be a
```

fullscreen.bs
```diff
@@ -436,7 +439,7 @@ could be an open <{dialog}> element.
 
  <p class=note>This is the set of documents for which the <a>fullscreen element</a> will be
  <a lt="unfullscreen an element">unfullscreened</a>, but the last document in <var>docs</var> might
- have more than one <a>element</a> in its <a>top layer</a> with the <a>fullscreen flag</a> set,
+ have more than one <a>element</a> in its [=Document/top layer=] with the <a>fullscreen flag</a> set,
```
foolip, line 442:
```suggestion
 have more than one <a>element</a> in its <a for=Document>top layer</a> with the <a>fullscreen flag</a> set,
```

fullscreen.bs
```diff
@@ -745,6 +661,13 @@ delivered with the <a>document</a> through which it is nested.
 <p>This prevents e.g. content from third parties to go fullscreen without explicit permission.
 
 
+<h2 id=old-links class=no-num oldids="new-stacking-layer, top-layer, top-layer-add, ::backdrop-pseudo-element, css-pe-backdrop">
+Appendix A: Previously-Hosted Definitions</h2>
+
+This specification previously hosted the definitions of <a selector>::backdrop</a>
+and the concept of the document's [=Document/top layer=].
```
foolip, line 668:
```suggestion
and the concept of the document's <a for=Document>top layer</a>.
```

fullscreen.bs
```diff
@@ -56,21 +57,22 @@ stated otherwise it is unset.
...
 <p>To <dfn>unfullscreen an <var>element</var></dfn>, unset <var>element</var>'s
-<a>fullscreen flag</a> and <a>iframe fullscreen flag</a> (if any), and <a for=set>remove</a> it from
-its <a>node document</a>'s <a>top layer</a>.
+<a>fullscreen flag</a> and <a>iframe fullscreen flag</a> (if any), and
+<a>remove from the top layer immediately</a> given |element|'s <a>node document</a> and |element|.
 
 <p>To <dfn>unfullscreen a <var>document</var></dfn>,
 <a lt="unfullscreen an element">unfullscreen</a> all <a>elements</a>, within <var>document</var>'s
-<a>top layer</a>, whose <a>fullscreen flag</a> is set.
+[=Document/top layer=], whose <a>fullscreen flag</a> is set.
```
foolip, line 75:
```suggestion
<a for=Document>top layer</a>, whose <a>fullscreen flag</a> is set.
```

foolip:
There are conflicts from https://github.com/whatwg/fullscreen/pull/204, I'll try to resolve...

fullscreen.bs
```diff
@@ -68,17 +69,18 @@ is an <a>ordered set</a> of (<a>string</a>, <a>element</a>) <a>tuples</a>. It is
 
  <li><p>Set <var>element</var>'s <a>fullscreen flag</a>.
 
- <li><p><a for="top layer">Add</a> <var>element</var> to <var>element</var>'s <a>node document</a>'s
- <a>top layer</a>.
+ <li><p><a>Request removal from the top layer</a> given |element|'s <a>node document</a> and |element|.
```
foolip, line 72:
```suggestion
 <li><p><a>Request removal from the top layer</a> given <var>element</var>.
```

fullscreen.bs
```diff
@@ -68,17 +69,18 @@ is an <a>ordered set</a> of (<a>string</a>, <a>element</a>) <a>tuples</a>. It is
...
+
+ <li><a>Add to the top layer</a> given |element|'s <a>node document</a> and |element|.
```
foolip, line 74:
```suggestion
 <li><a>Add to the top layer</a> given <var>element</var>.
```

fullscreen.bs
```diff
@@ -68,17 +69,18 @@ is an <a>ordered set</a> of (<a>string</a>, <a>element</a>) <a>tuples</a>. It is
...
 </ol>
 
 <p>To <dfn>unfullscreen an <var>element</var></dfn>, unset <var>element</var>'s
-<a>fullscreen flag</a> and <a>iframe fullscreen flag</a> (if any), and <a for=set>remove</a> it from
-its <a>node document</a>'s <a>top layer</a>.
+<a>fullscreen flag</a> and <a>iframe fullscreen flag</a> (if any), and
+<a>remove from the top layer immediately</a> given |element|'s <a>node document</a> and |element|.
```
foolip, line 79:
```suggestion
<a>remove from the top layer immediately</a> given <var>element</var>.
```

fullscreen.bs
```diff
@@ -117,10 +119,10 @@ these steps:
    <li><p>Otherwise, <a lt="unfullscreen an element">unfullscreen <var>node</var></a>.
 
    <li>
-    <p>If <var>document</var>'s <a>top layer</a> <a for=set>contains</a> <var>node</var>,
-    <a for=set>remove</a> <var>node</var> from <var>document</var>'s <a>top layer</a>.
+    <p>If <var>document</var>'s <a for=Document>top layer</a> <a for=set>contains</a> <var>node</var>,
+    <a>remove from the top layer immediately</a> given <var>document</var> and <var>node</var>.
```
foolip, line 123:
```suggestion
    <a>remove from the top layer immediately</a> given <var>node</var>.
```

foolip (commented):
Changes to adapt to https://github.com/w3c/csswg-drafts/commit/038516478f4cfb0b405291b67d845b77face09b5 and stick with `<var>` as the rest of the spec does.

foolip (approved):
This is all good now IMHO, much nicer with the dropped document argument to many algorithms.

@annevk can you review this from the WebKit perspective?

foolip (approved):
This is all good now IMHO, much nicer with the dropped document argument to many algorithms.

@annevk can you review this from the WebKit perspective?

fullscreen.bs
```diff
@@ -68,13 +71,14 @@ is an <a>ordered set</a> of (<a>string</a>, <a>element</a>) <a>tuples</a>. It is
 
  <li><p>Set <var>element</var>'s <a>fullscreen flag</a>.
 
- <li><p><a for="top layer">Add</a> <var>element</var> to <var>element</var>'s <a>node document</a>'s
- <a>top layer</a>.
+ <li><p><a>Request removal from the top layer</a> given <var>element</var>.
+
+ <li><a>Add to the top layer</a> given <var>element</var>.
```
annevk, line 76: Only the second step is needed here looking at the algorithm for "add".

annevk (commented):
I think all my concerns are editorial, but they would still be good to fix.

fullscreen.bs
foolip: Without the call to "request removal from the top layer", the assert would be hit, right?

fullscreen.bs
```diff
@@ -16,6 +16,9 @@ spec:dom
 spec:infra
     type:dfn; for:set; text:for each
     type:dfn; text:string
+spec:css-position-4
+    type:selector; text: ::backdrop
+    type:dfn; text:top layer
```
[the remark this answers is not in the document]
foolip: Yes, but the definitions from this spec are as well, so the terms are ambiguous. I think we can remove these after this PR has landed and the changes have flowed through.

fullscreen.bs
annevk: I see. So we preserve the behavior but allow for slightly novel behavior in new APIs. I guess that's okay. If we don't end up wanting new behavior though and I think @nt1m argued we didn't for perf reasons we should simplify this again.

Perhaps it's best to wait for @nt1m and merge this tomorrow. I can ping him.

tabatkins: I'm not sure what you mean by "allow for slightly novel behavior in new APIs". All APIs would cycle thru the algorithms in this same way, unless they wanted to get into the guts and directly manipulate the top layer list (which they shouldn't).

The only two behaviors that specs should end up landing on for "put something in the top layer when it's already there" are this (cycle it to the end of the list) and "fail", probably throwing an error when they do so.

fullscreen.bs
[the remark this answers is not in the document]
tabatkins: Yup, they're temporary additions to get the build green, and can be removed after all the updates land and the dfn sources are updated.

fullscreen.bs
nt1m, line 76: Should this be using immediate removal rather than requesting removal? Transitions wouldn't work here since we're adding it immediately back to the top layer.

nt1m (approved):
Looks fine aside from my one comment.

fullscreen.bs
tabatkins: It doesn't really matter which we use; they both produce identical behavior here.

nt1m: Reading the spec, it seems like they would but I don't see a reason to use the more complex one, when we can just use the simple one here.

annevk: Yeah, seems clearer too.

annevk: It does kinda seem like unless they would always remove prior to adding they'd have to perform a contains check. Which would be the somewhat novel behavior, but not entirely novel as the old API technically allowed for that as well.

foolip: Right, specs have to either check if the element is already in top layer and treat it as an error, or call "request removal from the top layer" to get the moving behavior. The latter preserves the behavior mentioned in the last note in https://fullscreen.spec.whatwg.org/commit-snapshots/afd56a35f409e5595dd861f41390a1016ebd6aa2/#new-stacking-layer.

foolip: OK, let's go with immediate removal.

foolip:
I've filed https://bugzilla.mozilla.org/show_bug.cgi?id=1834542 and https://bugs.webkit.org/show_bug.cgi?id=257199.

@annevk do you want to approve this?

annevk (approved):

annevk:
🎉  Thanks all!

PR #224, opened 2023-05-03
Flatten requestFullscreen() error conditions

foolip:
This is terrible. @annevk suggests a helper algorithm instead.

***
<a href="https://whatpr.org/fullscreen/224.html" title="Last updated on May 3, 2023, 7:58 AM UTC (1fb8b52)">Preview</a> | <a href="https://whatpr.org/fullscreen/224/b3bab96...1fb8b52.html" title="Last updated on May 3, 2023, 7:58 AM UTC (1fb8b52)">Diff</a>

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -246,25 +246,19 @@ are:
 
  <li><p>Let <var>error</var> be false.
 
- <li>
-  <p>If any of the following conditions are false, then set <var>error</var> to true:
+ <li><p>If <a>this</a>'s <a for=Element>namespace</a> is not the <a>HTML namespace</a> and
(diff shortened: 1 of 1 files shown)


annevk:
Yeah, something like

> Let _error_ be the result of 'is fullscreen prevented' given this.

would work I think. And is fullscreen prevented can return early so the checks stay simple, e.g.,

> If _element_ is a dialog element, then return true.

foolip:
I tried something in https://github.com/whatwg/fullscreen/compare/fullscreen-is-prevented?expand=1 but I don't really like that either. I'll just leave it on a branch in case I want to get back to it, and close this draft PR.

PR #232, opened 2023-12-06
Add keyboard locking option

### LINKED ISSUE #231: Keyboard lock
### What problem are you trying to solve?

Fullscreen applications might want to access keys that would ordinarily break out of the full screen state.  For instance, games often use the ESC key to access menus, remote desktop applications often use keyboard shortcuts for app switching.  Having these actions sent to the site, rather than the local system might be desirable.

### What solutions exist today?

Google previously proposed https://github.com/WICG/keyboard-lock/

### How would you solve it?

I made a proposal a long time ago here: https://docs.google.com/document/d/191YzTVJRg4mIkAYCHYbDYFugO4xiLw6xE_lDuUOfItQ/edit?usp=sharing

Some of those details might change in discussion.  For instance, I consider points made by others (in comments on that document) about system key lockout to be convincing.

### Anything else?

I'm making this contribution here to make clear that this proposal is covered by Mozilla's commitment to the WHATWG IPR policy.  My intention is not to drive this work, but am willing to support someone who wants to pursue it.

marcoscaceres:
Introduce a dictionary member `keyboardLock` to `requestFullscreen(options)` to request keyboard lock together with fullscreen.

Fixes #231 

   * [WebKit](https://github.com/WebKit/WebKit/commit/ceba7343025b0e47c2903877198c8bdd5aeda851)
   * Gecko
   * https://wpt.fyi/results/fullscreen/api?label=master&label=experimental&aligned&q=keyboard-lock 
   * https://github.com/web-platform-tests/wpt/pull/59726
   * Chromium: https://issues.chromium.org/issues/505427218
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=700123
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=313121
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: https://github.com/mdn/content/issues/43868
- [x] The top of this comment includes a [clear commit message](https://github.com/whatwg/meta/blob/main/COMMITTING.md) to use. 

***

### :boom: Error: 422 Unprocessable Entity :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on May 8, 2026, 9:36 AM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/publications/spec-generator/) - Spec Generator is the web service used to build bikeshed/ReSpec specs

:link: [Related URL](https://www.w3.org/publications/spec-generator/?type=bikeshed-spec&output=html&url=https%3A%2F%2Fraw.githubusercontent.com%2Fwhatwg%2Ffullscreen%2F252503e2787154decf9cdd9945cf07ebcc2c23f0%2Ffullscreen.bs&force=1&md-status=LS-PR&md-Text-Macro=PR-NUMBER%20232)

**Error output:**

```json
[
    {
        "lineNum": "114:35",
        "messageType": "link",
        "text": "Multiple possible 'removing steps' dfn refs.\nArbitrarily chose https://dom.spec.whatwg.org/#concept-node-remove-ext\nTo auto-select one of the following refs, insert one of these lines into a <pre class=link-defaults> block:\nspec:dom; type:dfn; text:removing steps\nspec:geolocation-element; type:dfn; for:ActivationBlockersMixin; text:removing steps\nspec:geolocation-element; type:dfn; for:HTMLGeolocationElement; text:removing steps\nspec:geolocation-element; type:dfn; for:PowerfulFeatureObserver; text:removing steps\n<a bs-line-number=\"114:35\" data-link-type=\"dfn\" data-lt=\"removing steps\">removing steps</a>"
    },
    {
        "lineNum": null,
        "messageType": "failure",
        "text": "Did not generate, due to errors exceeding the allowed error level."
    }
]
```

_This seems to be an issue with the [Spec Generator](https://www.w3.org/publications/spec-generator/) service. PR Preview doesn't manage this service and so has no control over it. If you've identified an issue with it, you can [report the issue to the maintainers of Spec Generator](https://github.com/w3c/spec-generator/issues/new) directly. Please be courteous. Thank you!_

_If you don't have enough information above to solve the error by yourself or if the issue doesn't seem related to Spec Generator, you can [file an issue with PR Preview](https://github.com/tobie/pr-preview/issues/new?title=Unidentified%20Error&body=See%20whatwg/fullscreen%23232.)._

</details>

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -195,7 +195,7 @@ enum FullscreenNavigationUI {
 };
 
 enum FullscreenKeyboardLock {
-  "application",
+  "browser",
   "none",
   "system"
 };
@@ -245,7 +245,7 @@ partial interface mixin DocumentOrShadowRoot {
   When supplied, <var>options</var>'s {{FullscreenOptions/keyboardLock}} member indicates whether
   the supplied keyboard lock mode should apply.
 
-  If set to "{{FullscreenKeyboardLock/application}}", [=application-level keyboard lock=] should
+  If set to "{{FullscreenKeyboardLock/browser}}", [=application-level keyboard lock=] should
   be applied.
 
   If set to "{{FullscreenKeyboardLock/system}}", [=system-level keyboard lock=] should be applied.
@@ -423,7 +423,7 @@ are:
   <li>
   Depending on the value of {{FullscreenOptions}}'s {{FullscreenOptions/keyboardLock}}:
   <dl switch>
-    <dt>"{{FullscreenKeyboardLock/application}}"</dt>
+    <dt>"{{FullscreenKeyboardLock/browser}}"</dt>
     <dd><p>Apply an [=application-level keyboard lock=].
     <dt>"{{FullscreenKeyboardLock/system}}"</dt>
     <dd><p>Apply a [=system-level keyboard lock=].
@@ -709,7 +709,7 @@ like ESC or function keys are integral to the application's functionality.
 
 <p>A <dfn>keyboard lock</dfn> enables web applications to capture and handle key inputs directly,
 bypassing the default behavior typically executed by the user agent (in the case of
-"{{FullscreenKeyboardLock/application}}") or operating system (in the case of
+"{{FullscreenKeyboardLock/browser}}") or operating system (in the case of
 "{{FullscreenKeyboardLock/system}}"). When the [=keyboard lock flag=] is set for a [=document=], key
 events that would normally trigger user agent or system-level actions are instead redirected to the
 web application in fullscreen.
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -376,7 +376,7 @@ are:
   </ol>
 
  <li>
-  If, {{FullscreenOptions}}'s {{FullscreenOptions/keyboardLock}}'s value is not supported,
+  If {{FullscreenOptions}}'s {{FullscreenOptions/keyboardLock}}'s value is not supported,
   reject <var>promise</var> with a {{NotSupportedError}} exception and terminate these
   steps.
  </li>
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -432,10 +432,10 @@ are:
 
   <dl class=switch>
    <dt>"{{FullscreenKeyboardLock/browser}}"</dt>
-   <dd><p>Set the [=keyboard lock flag=] for <a>this</a> to true.
+   <dd><p>Se <a>this</a>'s [=keyboard lock flag=].
 
    <dt>"{{FullscreenKeyboardLock/none}}"</dt>
-   <dd><p>Set the [=keyboard lock flag=] for <a>this</a> to false.
+   <dd><p>Unset <a>this</a>'s [=keyboard lock flag=].
   </dl>
 
  <li><p>Resolve <var>promise</var> with undefined.
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -87,9 +87,9 @@ is an <a>ordered set</a> of (<a>string</a>, <a>element</a>) <a>tuples</a>. It is
  <li><a>Add to the top layer</a> given <var>element</var>.
 </ol>
 
-<p>To <dfn>unfullscreen an <var>element</var></dfn>, unset <var>element</var>'s
-<a>fullscreen flag</a>, <a>iframe fullscreen flag</a> (if any), and <a>keyboard lock flag</a>, and
-<a>remove from the top layer immediately</a> given <var>element</var>.
+<p>To <dfn>unfullscreen an <var>element</var></dfn>, unset <var>element</var>'s <a>fullscreen
+flag</a>, <a>iframe fullscreen flag</a> (if any), and <a>keyboard lock flag</a>, and <a>remove from
+the top layer immediately</a> given <var>element</var>.
 
 <p>To <dfn>unfullscreen a <var>document</var></dfn>,
 <a lt="unfullscreen an element">unfullscreen</a> all <a>elements</a>, within <var>document</var>'s
@@ -180,9 +180,11 @@ steps:
 
 <hr>
 
-<p>All <a>elements</a> have an associated <dfn>keyboard lock flag</dfn>. Unless stated otherwise it is unset.
+<p>All <a>elements</a> have an associated <dfn>keyboard lock flag</dfn>. Unless stated otherwise it
+is unset.
 
-<p>A [=document=] has <dfn>keyboard lock</dfn> active when its [=fullscreen element=] is not null and the [=fullscreen element=]'s [=keyboard lock flag=] is true, and inactive otherwise.
+<p>A [=document=] has <dfn>keyboard lock</dfn> active when its [=fullscreen element=] is not null
+and the [=fullscreen element=]'s [=keyboard lock flag=] is true, and inactive otherwise.
 
 <div algorithm>
   <p>To <dfn>run the release the keyboard lock steps</dfn> for a <a>document</a>
@@ -191,9 +193,11 @@ steps:
   <ol>
     <li><p>If <var>document</var>'s <a>fullscreen element</a> is null, then return.
 
-    <li><p>If the [=keyboard lock flag=] for <var>document</var>'s <a>fullscreen element</a> is false, then return.
+    <li><p>If the [=keyboard lock flag=] for <var>document</var>'s <a>fullscreen element</a> is
+    false, then return.
 
-    <li><p>Set the [=keyboard lock flag=] for <var>document</var>'s <a>fullscreen element</a> to false.
+    <li><p>Set the [=keyboard lock flag=] for <var>document</var>'s <a>fullscreen element</a> to
+    false.
   </ol>
 </div>
 
@@ -716,7 +720,9 @@ fullscreen. Such key events (for individual keys or keyboard shortcuts) may eith
 while [=keyboard lock=] is active, or it could still have the same action but allow the web page to
 call {{Event/preventDefault()}} to cancel the action.
 
-<p>Whenever a [=document=]'s [=keyboard lock=] is changed from active to inactive, user agents must deactivate the keyboard lock and restore the handling of keyboard inputs to the default behavior of the user agent and the operating system.
+<p>Whenever a [=document=]'s [=keyboard lock=] is changed from active to inactive, user agents must
+deactivate the keyboard lock and restore the handling of keyboard inputs to the default behavior of
+the user agent and the operating system.
 
 <p>User agents should reserve an additional input for the purposes of exiting fullscreen. There are
 also some system-level key sequences or combinations that cannot be intercepted for security
@@ -766,7 +772,11 @@ system control or privacy risks (like Alt+Tab, Ctrl+Alt+Del), can be locked via 
 <p>When applying a [=keyboard lock=], user agents should implement safeguards to prevent abuse of
 this feature, such as allowing users to override the lock.
 
-<p>User agents may provide a user preference to ignore [=keyboard lock=] (with any granularity, e.g. for a specific site, or globally). Similarly, some platforms might not have a keyboard, where user agents may ignore the [=keyboard lock=] state. However, this should not affect the web-observable behavior of the {{Element/requestFullscreen()}} method or be exposed in other ways, to avoid fingerprinting.
+<p>User agents may provide a user preference to ignore [=keyboard lock=] (with any granularity, e.g.
+for a specific site, or globally). Similarly, some platforms might not have a keyboard, where user
+agents may ignore the [=keyboard lock=] state. However, this should not affect the web-observable
+behavior of the {{Element/requestFullscreen()}} method or be exposed in other ways, to avoid
+fingerprinting.
 
 
 
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -16,6 +16,7 @@ spec:dom
     type:dfn; for:/; text:document
     type:dfn; for:/; text:element
     type:interface; text:Document
+    type:dfn; text:removing steps
 spec:infra
     type:dfn; for:set; text:for each
     type:dfn; text:string
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -184,20 +184,19 @@ steps:
 is unset.
 
 <p>A [=document=] has <dfn>keyboard lock</dfn> active when its [=fullscreen element=] is not null
-and the [=fullscreen element=]'s [=keyboard lock flag=] is true, and inactive otherwise.
+and the [=fullscreen element=]'s [=keyboard lock flag=] is set, and inactive otherwise.
 
 <div algorithm>
-  <p>To <dfn>run the release the keyboard lock steps</dfn> for a <a>document</a>
-  <var>document</var>, run these steps:
+  <p>To <dfn>release the keyboard lock</dfn> for a <a>document</a> <var>document</var>, run these
+  steps:
 
   <ol>
     <li><p>If <var>document</var>'s <a>fullscreen element</a> is null, then return.
 
-    <li><p>If the [=keyboard lock flag=] for <var>document</var>'s <a>fullscreen element</a> is
-    false, then return.
+    <li><p>If <var>document</var>'s <a>fullscreen element</a>'s [=keyboard lock flag=] is unset,
+    then return.
 
-    <li><p>Set the [=keyboard lock flag=] for <var>document</var>'s <a>fullscreen element</a> to
-    false.
+    <li><p>Unset <var>document</var>'s <a>fullscreen element</a>'s [=keyboard lock flag=].
   </ol>
 </div>
 
@@ -553,7 +552,7 @@ could be an open <{dialog}> element.
 
  <li><p>Return <var>promise</var>, and run the remaining steps <a>in parallel</a>.
 
- <li><p>[=Run the release the keyboard lock steps=] with <var>doc</var>.
+ <li><p>[=Release the keyboard lock=] for <var>doc</var>.
 
  <li><p>Run the [=fully unlock the screen orientation steps=] with <var>doc</var>.
 
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -183,8 +183,8 @@ steps:
 <p>All <a>elements</a> have an associated <dfn>keyboard lock flag</dfn>. Unless stated otherwise it
 is unset.
 
-<p>A [=document=] has <dfn>keyboard lock</dfn> active when its [=fullscreen element=] is not null
-and the [=fullscreen element=]'s [=keyboard lock flag=] is set, and inactive otherwise.
+<p>A <a>document</a> has <dfn>keyboard lock</dfn> active when its <a>fullscreen element</a> is not
+null and the <a>fullscreen element</a>'s <a>keyboard lock flag</a> is set, and inactive otherwise.
 
 <div algorithm>
   <p>To <dfn>release the keyboard lock</dfn> for a <a>document</a> <var>document</var>, run these
@@ -193,10 +193,10 @@ and the [=fullscreen element=]'s [=keyboard lock flag=] is set, and inactive oth
   <ol>
     <li><p>If <var>document</var>'s <a>fullscreen element</a> is null, then return.
 
-    <li><p>If <var>document</var>'s <a>fullscreen element</a>'s [=keyboard lock flag=] is unset,
+    <li><p>If <var>document</var>'s <a>fullscreen element</a>'s <a>keyboard lock flag</a> is unset,
     then return.
 
-    <li><p>Unset <var>document</var>'s <a>fullscreen element</a>'s [=keyboard lock flag=].
+    <li><p>Unset <var>document</var>'s <a>fullscreen element</a>'s <a>keyboard lock flag</a>.
   </ol>
 </div>
 
@@ -258,7 +258,7 @@ partial interface mixin DocumentOrShadowRoot {
   When supplied, <var>options</var>'s {{FullscreenOptions/keyboardLock}} member indicates whether
   the supplied keyboard lock mode will apply.
 
-  If set to "{{FullscreenKeyboardLock/browser}}", [=keyboard lock=] can be applied.
+  If set to "{{FullscreenKeyboardLock/browser}}", <a>keyboard lock</a> can be applied.
 
   User agents are always free to honor user preference over the application's. The default value
   "{{FullscreenKeyboardLock/none}}" indicates no keyboard lock is applied.
@@ -432,10 +432,10 @@ are:
 
   <dl class=switch>
    <dt>"{{FullscreenKeyboardLock/browser}}"</dt>
-   <dd><p>Set <a>this</a>'s [=keyboard lock flag=].
+   <dd><p>Set <a>this</a>'s <a>keyboard lock flag</a>.
 
    <dt>"{{FullscreenKeyboardLock/none}}"</dt>
-   <dd><p>Unset <a>this</a>'s [=keyboard lock flag=].
+   <dd><p>Unset <a>this</a>'s <a>keyboard lock flag</a>.
   </dl>
 
  <li><p>Resolve <var>promise</var> with undefined.
@@ -552,9 +552,9 @@ could be an open <{dialog}> element.
 
  <li><p>Return <var>promise</var>, and run the remaining steps <a>in parallel</a>.
 
- <li><p>[=Release the keyboard lock=] for <var>doc</var>.
+ <li><p><a>Release the keyboard lock</a> for <var>doc</var>.
 
- <li><p>Run the [=fully unlock the screen orientation steps=] with <var>doc</var>.
+ <li><p>Run the <a>fully unlock the screen orientation steps</a> with <var>doc</var>.
 
  <li><p>If <var>resize</var> is true, resize <var>doc</var>'s viewport to its "normal" dimensions.
 
@@ -643,8 +643,8 @@ algorithm as part of the <a>close request steps</a>. This takes precedence over 
 <p>The user agent may end any fullscreen session without a <a>close request</a> or call to
 {{Document/exitFullscreen()}} whenever the user agent deems it necessary.
 
-<p>Users should be clearly notified when [=keyboard locking=] is active, possibly through browser UI
-indicators.
+<p>Users should be clearly notified when <a>keyboard locking</a> is active, possibly through browser
+UI indicators.
 
 <p>There should be a simple and intuitive method for users to override keyboard locking, reverting
 control back to the system or user agent.
@@ -712,16 +712,16 @@ them to intercept certain key inputs that would typically be handled by either t
 user agent. This is useful in scenarios such as gaming or remote desktop applications, where keys
 like ESC or function keys are integral to the application's functionality.
 
-<p>A [=keyboard lock=] enables web applications to capture and handle key inputs directly, bypassing
-the default behavior typically executed by the user agent or operating system. Key events that would
-normally trigger user agent or system-level actions are instead redirected to the web application in
-fullscreen. Such key events (for individual keys or keyboard shortcuts) may either have no action
-while [=keyboard lock=] is active, or it could still have the same action but allow the web page to
-call {{Event/preventDefault()}} to cancel the action.
+<p>A <a>keyboard lock</a> enables web applications to capture and handle key inputs directly,
+bypassing the default behavior typically executed by the user agent or operating system. Key events
+that would normally trigger user agent or system-level actions are instead redirected to the web
+application in fullscreen. Such key events (for individual keys or keyboard shortcuts) may either
+have no action while <a>keyboard lock</a> is active, or it could still have the same action but
+allow the web page to call {{Event/preventDefault()}} to cancel the action.
 
-<p>Whenever a [=document=]'s [=keyboard lock=] is changed from active to inactive, user agents must
-deactivate the keyboard lock and restore the handling of keyboard inputs to the default behavior of
-the user agent and the operating system.
+<p>Whenever a <a>document</a>'s <a>keyboard lock</a> is changed from active to inactive, user agents
+must deactivate the keyboard lock and restore the handling of keyboard inputs to the default
+behavior of the user agent and the operating system.
 
 <p>User agents should reserve an additional input for the purposes of exiting fullscreen. There are
 also some system-level key sequences or combinations that cannot be intercepted for security
@@ -765,17 +765,17 @@ delivered with the <a>document</a> through which it is nested.
 
 <p>This prevents e.g. content from third parties to go fullscreen without explicit permission.
 
-<p>When applying a [=keyboard lock=], only a limited set of keys, primarily those not involving
+<p>When applying a <a>keyboard lock</a>, only a limited set of keys, primarily those not involving
 system control or privacy risks (like Alt+Tab, Ctrl+Alt+Del), can be locked via the API.
 
-<p>When applying a [=keyboard lock=], user agents should implement safeguards to prevent abuse of
+<p>When applying a <a>keyboard lock</a>, user agents should implement safeguards to prevent abuse of
 this feature, such as allowing users to override the lock.
 
-<p>User agents may provide a user preference to ignore [=keyboard lock=] (with any granularity, e.g.
-for a specific site, or globally). Similarly, some platforms might not have a keyboard, where user
-agents may ignore the [=keyboard lock=] state. However, this should not affect the web-observable
-behavior of the {{Element/requestFullscreen()}} method or be exposed in other ways, to avoid
-fingerprinting.
+<p>User agents may provide a user preference to ignore <a>keyboard lock</a> (with any granularity,
+e.g. for a specific site, or globally). Similarly, some platforms might not have a keyboard, where
+user agents may ignore the <a>keyboard lock</a> state. However, this should not affect the
+web-observable behavior of the {{Element/requestFullscreen()}} method or be exposed in other ways,
+to avoid fingerprinting.
 
 
 
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -64,9 +64,9 @@ stated otherwise it is unset.
 
 <p>All <a for=/>documents</a> have an associated <dfn>keyboard lock flag</dfn>. This flag indicates
 whether the keyboard lock is active for the document. When the flag is set, the document captures
-and handles key inputs as specified by a passed {{FullscreenKeyboardLock}} enum value. When the
-flag is not set, keyboard inputs are processed in the default manner by the user agent and the
-operating system. Unless stated otherwise it is unset.
+and handles key inputs as specified by a passed {{FullscreenKeyboardLock}} enum value. When the flag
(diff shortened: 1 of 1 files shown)


fullscreen.bs
```diff
@@ -640,6 +688,40 @@ iframe:fullscreen {
...
+"{{FullscreenKeyboardLock/system}}"). When the [=keyboard lock flag=] is set for a [=document=], key
+events that would normally trigger user agent or system-level actions are instead redirected to the
+web application in fullscreen.
+
+<dl>
+  <dt><dfn>Application-level keyboard lock</dfn></dt>
+  <dd><p>Captures all keys that the user agent would ordinarily receive, except those reserved for
+  critical system functions or security shortcuts (e.g., Ctrl+Alt+Del).
+  <dt><dfn>System-level keyboard lock</dfn></dt>
+  <dd><p>capture a wider range of keys, including those used for system-level navigation and
```
marcoscaceres, line 710:
```suggestion
  <dd><p>Capture a wider range of keys, including those used for system-level navigation and
```

fullscreen.bs
```diff
@@ -57,6 +57,12 @@ unset.
 <p>All <{iframe}> <a>elements</a> have an associated <dfn>iframe fullscreen flag</dfn>. Unless
 stated otherwise it is unset.
 
+<p>All <a for=/>documents</a> have an associated <dfn>keyboard lock flag</dfn>. This flag indicates
+whether the keyboard lock is active for the document. When the flag is set, the document captures
+and handles key inputs as specified by a passed {{FullscreenKeyboardLock}} enum value. When the
+flag is not set, keyboard inputs are processed in the default manner by the user agent and the
+operating system.
```
marcoscaceres, line 64:
```suggestion
operating system. Unless stated otherwise it is unset.
```

martinthomson (commented):
Pretty good. Some suggestions.

fullscreen.bs
```diff
@@ -583,6 +626,11 @@ corresponding <a>event handler content attributes</a> for {{Element}} objects in
 <p>The user agent may end any fullscreen session without instruction from the end user
 or call to {{Document/exitFullscreen()}} whenever the user agent deems it necessary.
 
+<p>Users should be clearly notified when [=keyboard locking=] is active, possibly through browser
+UI indicators.
+
+<p>There should be a simple and intuitive method for users to override keyboard locking, reverting
```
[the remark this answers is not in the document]
marcoscaceres: Agree... though this might need to happen after browsers ship something and if we all align. I guess it would be a SHOULD (or even MAY) level thing. History suggests that we will naturally converge on a UI/UX convention for this.

fullscreen.bs
```diff
@@ -178,7 +184,14 @@ enum FullscreenNavigationUI {
   "hide"
 };
 
+enum FullscreenKeyboardLock {
+  "application",
```
mdevaev: It's probably worth renaming it to `browser`. [WebKit's implementation](https://github.com/WebKit/WebKit/pull/21374/files#diff-8b54e72a7adef28afc726456ac23fc80c68f4d870a1a37ddadf9ad5e22df1107R38) of the [original proposal](https://docs.google.com/document/d/191YzTVJRg4mIkAYCHYbDYFugO4xiLw6xE_lDuUOfItQ/edit?tab=t.0#) has uses a `browser` there, so it would be great to avoid confusion.

mdevaev:
Hello. Please forgive me for impudently and unceremoniously barging in :)

WebKit has recently received an implementation of Mozilla's proposal, and apparently at some point it will reach Safari and other browsers: https://bugs.webkit.org/show_bug.cgi?id=264843

As a developer of a web-based for remote access device, I'm super-interested in this feature because it's the only blocker that prevents me from recommending Firefox instead of Chrome to my clients.

I hope that if this specification is adopted, then eventually things will get off the ground, and it will be done in Firefox too. Considering that we already have an actual implementation, and if there are no problems here, is it worth merging this PR? It looks great.

lukewarlow:
Per https://webkit.org/blog/17674/release-notes-for-safari-technology-preview-234/ I believe this has shipped in Safari TP. It would be good to check how accurate the spec PR is compared to the implementation and try to get it across the line now there's some implementation interest.

marcoscaceres:
Ok, had a look. [the WebKit PR](https://github.com/WebKit/WebKit/commit/ceba7343025b0e47c2903877198c8bdd5aeda851) does follow the spec. The only thing is that the tests that were added use WebKit infrastructure, but I think they could be migrated to be fullscreen tests... assuming we can lock and we all generally agree that hitting "ESC" once shouldn't cause fullscreen to cancel if the lock if applied (this might be hard to test across UAs... but we can try).  

In the meantime we need interest from a second implementer.

mdevaev:
As a developer, I would like to note that there is exactly one problem with the current draft/implementation: it is impossible to check the presence or absence of `keyboardLock` support on the target browser without having to call `requestFullscreen()`. There is no any flag or property that we could use for this. And even after the call, the code does not know whether it intercepts all key events.

If this is not supported, it would be great to be able to distinguish between these states to alert the user who expects all his keys to be intercepted.

It might not be too late to add something for that.

fullscreen.bs
```diff
@@ -178,7 +184,14 @@ enum FullscreenNavigationUI {
...
+  "none",
+  "system"
+};
+
```
saschanaz, line 202: This is not what [STP is shipping now](https://searchfox.org/wubkat/rev/6e0c8e30ad2f1f8915549b6746fc43fa9c7f5a87/Source/WebCore/dom/FullscreenOptions.idl#37-39), are we going to update this or is WebKit going to change?

fullscreen.bs
```diff
@@ -640,6 +688,46 @@ iframe:fullscreen {
...
+web application in fullscreen.
+
+<dl>
+  <dt><dfn>Application-level keyboard lock</dfn></dt>
+  <dd><p>Captures all keys that the user agent would ordinarily receive, such as those that open or
+  close windows. This excludes those reserved for system-level interactions, such
+  as key presses that open system menus or switch applications.
+  <dt><dfn>System-level keyboard lock</dfn></dt>
+  <dd><p>Capture a wider range of keys, including those used for system-level navigation and
+  shortcuts (e.g., Alt+Tab), subject to user consent and user agent implementation.
```
saschanaz: I think STP is not doing this (I don't see any non-browser shortcut is being captured), do we need this in that case?

Also, can environments like Linux snap/flatpak or Windows sandboxed MSIX support this? Sounds like we would surprisingly expose browser sandbox status if not.

saschanaz (changes requested):
based on above.

saschanaz:
I'm concerned that Safari would ship with something nobody agreed on, without an updated PR. If @marcoscaceres is busy I can take this and update it. (In that case I'll exclude system lock mode.)

fullscreen.bs
```diff
@@ -178,7 +184,14 @@ enum FullscreenNavigationUI {
...
 };
 
+enum FullscreenKeyboardLock {
+  "application",
+  "none",
+  "system"
+};
+
 dictionary FullscreenOptions {
+  FullscreenKeyboardLock keyboardLock = "none";
```
saschanaz, line 204: But do we need "none"? Maybe `undefined` can serve that purpose?

saschanaz:
I'm starting to think that all we need for `keyboardLock: "browser"` is to make the escape key not immediately escape fullscreen. Do we even need to call it "keyboard lock" in that case?

(Because browser shortcuts are already preventDefault-able.)

martinthomson:
Are you suggesting that sites looking to avoid things like tab switching have some enumeration of all browser shortcuts and just `.preventDefault()` on all of them?  Isn't there at least some value in having that done for them?

saschanaz:
> Are you suggesting that sites looking to avoid things like tab switching have some enumeration of all browser shortcuts and just `.preventDefault()` on all of them? Isn't there at least some value in having that done for them?

(TIL Firefox and Chrome even allow that...)

But anyway that's what Safari is going to ship now - there's no "bypass" as this PR says, all it does is to prevent the escape key to not immediately escape the fullscreen mode. There's no even extra capture for shortcuts, it turns out that both Safari and Chrome already allow the page to capture extra browser shortcuts on fullscreen mode [that Firefox currently does not](https://bugzilla.mozilla.org/show_bug.cgi?id=1593883).

And I do think it may indeed be easier to just prevent-default those shortcuts in keyboard lock mode, but I'm not sure preventing all shortcuts is a great idea. Is Control+C a browser shortcut to prevent, for example? What is the definition of "browser shortcuts that we want to prevent" here?

fullscreen.bs
marcoscaceres: Maybe... what I like about "none" is that it's explicit about "no lock is applied"... vs `undefined`, which, yes, could mean "no lock is applied", but also means "undefined behavior (and potentially the browser is applying UA specific behavior)". 

My question is, which one do we want? These are subtle, but important differences IMO,

saschanaz: But `keyboardLock: undefined` behaves exactly like "none" already?

fullscreen.bs
zcorpan: Per https://developer.apple.com/documentation/safari-release-notes/safari-26_4-release-notes#New-Features:~:text=Added%20support%20for%20the%20Keyboard%20Lock%20API%2E%20%28161422221 it seems this has now shipped in Safari stable.

The implemented keyword is still "browser", not "application". I think the spec should be changed to match.

fullscreen.bs
zcorpan: `undefined` should be equivalent to `"none"`, not "undefined behavior". Whether to include `"none"` is a question of API design, though I don't see harm in including it as its own keyword.

zcorpan:
Since Safari has shipped, the cost of tweaking the API is higher now. I think we should keep the "none" and "browser" enums, but drop "system".

It should be up to the UA to decide which keys are reserved with keyboard lock active (and when not active).

zcorpan:
Adding Gecko as interested implementer (see https://bugzilla.mozilla.org/show_bug.cgi?id=700123).

zcorpan:
> As a developer, I would like to note that there is exactly one problem with the current draft/implementation: it is impossible to check the presence or absence of `keyboardLock` support on the target browser without having to call `requestFullscreen()`. There is no any flag or property that we could use for this. And even after the call, the code does not know whether it intercepts all key events.
> 
> If this is not supported, it would be great to be able to distinguish between these states to alert the user who expects all his keys to be intercepted.
> 
> It might not be too late to add something for that.

It's possible to feature-check that the `keyboardLock` dictionary member is supported without going fullscreen like this: https://software.hixie.ch/utilities/js/live-dom-viewer/saved/14660

This doesn't tell you whether a particular value is supported until actually going fullscreen, though. This step:

> If FullscreenOptions 's[ keyboardLock ](https://whatpr.org/fullscreen/232/2263e15...f670402.html#dom-fullscreenoptions-keyboardlock)'s value is not supported, reject promise with a NotSupportedError exception and terminate these steps.

Runs after rejecting when |error| is true, so you can only tell that the value is supported by going fullscreen with requesting `keyboardLock` (and checking that the dictionary member is supported first, since unknown dictionary members are silently ignored).

While this method of checking dictionary members has been deemed sufficient for feature-checking in some cases, I could be convinced that it's not ideal DX and there should be a property similar to `document.fullscreenEnabled` (e.g. `document.keyboardLockEnabled`) to allow easy feature-checking and easy checking if keyboard lock is currently active. (If we drop `system` then this becomes a boolean.)

mdevaev:
> I could be convinced that it's not ideal DX and there should be a property similar to `document.fullscreenEnabled` (e.g. `document.keyboardLockEnabled`) to allow easy feature-checking and easy checking if keyboard lock is currently active. (If we drop `system` then this becomes a boolean.)

Thanks for the previous example. I also think that more explicit feature testing is needed, otherwise it is not very clear how to understand which keys are intercepted and which are not (at least in the form of the name of the set), and convey this information to the user.

zcorpan:
I've removed the "system" enum and renamed "application" to "browser".

mdevaev:
> I've removed the "system" enum

Why? If you want to completely eliminate the possibility of this, then I can give you a good use case when necessary: if you are making a remote access client (something like VNC or TeamViewer) and want the maximum possible number of keys and their combinations to be intercepted. I have just such a case: users want the keys to be able to be transferred to a remote side.

zcorpan:
The idea is to provide the maximum possible number of keys (but up to the UA to decide) in the "browser" mode.

mdevaev:
Wouldn't this conflict with Safari's implementation? Plus, it's a bit counterintuitive that "browser" means "all possible hotkeys."

zcorpan:
No, in Safari "system" is the same as "none". As far as I can tell, the only difference between "none" and "browser" in Safari is that the Escape key is `preventDefault()`-able.

zcorpan:
I think having the keyboard lock flag on `Document` doesn't work well, because a document can have multiple fullscreened elements, and `document.exitFullscreen()` only unfullscreens the fullscreen element, it doesn't fully exit fullscreen. Elements have a [fullscreen flag](https://fullscreen.spec.whatwg.org/#fullscreen-flag). If the keyboard lock flag is also on elements, it's possible to set the keyboard lock state correctly when unfullscreening an element in the multiple fullscreened elements case.

Demo: https://software.hixie.ch/utilities/js/live-dom-viewer/saved/14667

hamishwillee:
FF151 adds support for this interface as per this spec. I'm supposed to document it - tracking bug is https://github.com/mdn/content/issues/43868

However this update hasn't been accepted, and there seems to be some questions about the implementation in the discussion. As such the  implementations will be "non-standard" which makes documenting from a baseline hard.

Can you help me by advising/confiming on the following?

1. From https://github.com/whatwg/fullscreen/pull/232/changes#r2754161239 Safari seems to have an extra mode option `system`, but is otherwise the same.
   - Do `none` and `browser` on Safari mean the same thing as in the spec.
   - What does "system" mode mean/how does it differ from `browser`.
2. The update indicates that browsers should provide alternative mechanisms for exiting fullscreen.
   - AIUI the methods are not mandated right? 
   - The FF implementation uses long-press Escape. Do we know if Safari and Chrome use or intend to use long-press escape too as the exit-fullscreen key? 
     - If so, would that likely become mandated by the spec?
   - What is the broad thinking for Android - we use `browser` mode for that, but some other key can be the escape key? 
3. The spec says:
   > may either have no action while [=keyboard lock=] is active, or it could still have the same action but allow the web page to call {{Event/preventDefault()}} to cancel the action.

   Doesn't that mean that a web application will always have to call `preventDefault()` to use a key that might be used by something else normally - because it can't know from whether the event default action has been disabled or not?

4. Firefox release note states of the implementation "This prevents the <kbd>Esc</kbd> key from causing the element to leave fullscreen (instead, a long-press is required), and some formerly-reserved browser hotkeys are allowed to be default-prevented."

   - Is the Esc key change "standard"? I mean do all browsers use Esc normally for leaving fullscreen mode (is that in the spec?) . I guess this is really back to point 3, trying to give users a flavour of what they need to call `preventDefault()` for, and what is done for them.

marcoscaceres:
@zcorpan, @saschanaz, should we have a call and sort this whole thing out? We probably need an hour just to take stock and figure out a path forward.

zcorpan:
Happy to have a call.

> FF151 adds support for this interface as per this spec. I'm supposed to document it - tracking bug is [mdn/content#43868](https://github.com/mdn/content/issues/43868)
> 
> However this update hasn't been accepted, and there seems to be some questions about the implementation in the discussion. As such the implementations will be "non-standard" which makes documenting from a baseline hard.

I would expect this can land pretty soon, though.

> Can you help me by advising/confiming on the following?
> 
>   1. From https://github.com/whatwg/fullscreen/pull/232/changes#r2754161239 Safari seems to have an extra mode option `system`, but is otherwise the same.
>        
>   * Do `none` and `browser` on Safari mean the same thing as in the spec.

I think so.

>    * What does "system" mode mean/how does it differ from `browser`.

From what I can tell, it doesn't do anything in WebKit (so is the same as `none`).

>   2. The update indicates that browsers should provide alternative mechanisms for exiting fullscreen.
>        
>    * AIUI the methods are not mandated right?

"should" is a normative requirement, see RFC 2119. The specific alternative mechanisms are up to the UA.

>    * The FF implementation uses long-press Escape. Do we know if Safari and Chrome use or intend to use long-press escape too as the exit-fullscreen key?

In Chrome (with `navigator.keyboard.lock(); el.requestFullscreen()`), long-press esc (1500ms in my testing) exits fullscreen.

In Safari with `el.requestFullscreen({keybaordLock: "browser"})`, long-press esc (500ms in my testing) exits fullscreen.
          
>   * If so, would that likely become mandated by the spec?

It's not mandated currently in this PR.

>    * What is the broad thinking for Android - we use `browser` mode for that, but some other key can be the escape key?

I think there's not the same need for keyboard lock on Android, so possibly it could be equivalent to "none".

> 
>   3. The spec says:
>      > may either have no action while [=keyboard lock=] is active, or it could still have the same action but allow the web page to call {{Event/preventDefault()}} to cancel the action.
>        
>        
>      Doesn't that mean that a web application will always have to call `preventDefault()` to use a key that might be used by something else normally - because it can't know from whether the event default action has been disabled or not?

Yes.

We could maybe standardize that the esc key should have no action when keyboard lock is active.

 
>   4. Firefox release note states of the implementation "This prevents the Esc key from causing the element to leave fullscreen (instead, a long-press is required), and some formerly-reserved browser hotkeys are allowed to be default-prevented."
>        
>      * Is the Esc key change "standard"? I mean do all browsers use Esc normally for leaving fullscreen mode (is that in the spec?) . I guess this is really back to point 3, trying to give users a flavour of what they need to call `preventDefault()` for, and what is done for them.

All desktop browsers as far as I know use the esc key for exiting fullscreen. In practice, Safari currently requires `preventDefault()` for the esc key when keyboard lock is active.

zcorpan:
> I think having the keyboard lock flag on `Document` doesn't work well, because a document can have multiple fullscreened elements, and `document.exitFullscreen()` only unfullscreens the fullscreen element, it doesn't fully exit fullscreen. Elements have a [fullscreen flag](https://fullscreen.spec.whatwg.org/#fullscreen-flag). If the keyboard lock flag is also on elements, it's possible to set the keyboard lock state correctly when unfullscreening an element in the multiple fullscreened elements case.
> 
> Demo: https://software.hixie.ch/utilities/js/live-dom-viewer/saved/14667

Fixed in 11a7f57.

zcorpan:
In https://github.com/whatwg/fullscreen/pull/232/commits/f4149b1c6a26917a0c472f64f149c52f17cc0e05 I changed what happens when keyboard lock is not supported (by the platform or for security reasons etc); instead of rejecting with `NotSupportedError` it follows the same steps as other error cases (so no transient activation is consumed, a `fullscreenerror` event is fired, and rejects with `TypeError`).

However maybe it should still successfully fullscreen in this case, just without keyboard lock. We can add `document.keyboardLockEnabled` and `document.keyboardLock` to mirror `document.fullscreenEnabled` and `document.fullscreen`, so it's easy to tell if it will be possible to acquire keyboard lock and whether it's currently active.

saschanaz:
`document.fullscreen` is deprecated though.

https://fullscreen.spec.whatwg.org/#dom-document-fullscreen

> Use the [fullscreenElement](https://fullscreen.spec.whatwg.org/#dom-document-fullscreenelement) attribute instead.

I wonder we can convert it to a dictionary that represents fullscreen options.

re: keyboardLockEnabled, that effectively sounds like it will expose keyboard existence.

zcorpan:
> `document.fullscreen` is deprecated though.
> 
> https://fullscreen.spec.whatwg.org/#dom-document-fullscreen

Yes, because `fullscreenElement` exists, but I think a boolean seems more useful for keyboard lock.

> > Use the [fullscreenElement](https://fullscreen.spec.whatwg.org/#dom-document-fullscreenelement) attribute instead.
> 
> I wonder we can convert it to a dictionary that represents fullscreen options.

Why? `navigationUI` seems like it doesn't need to be exposed.

> re: keyboardLockEnabled, that effectively sounds like it will expose keyboard existence.

Good point, maybe the API should act like it's always available even if there's no keyboard. Then I guess we also don't want to reject in `requestFullscreen`.

saschanaz:
> Why? `navigationUI` seems like it doesn't need to be exposed.

Because I want it to be future-ready, so that if we add anything more in the options bag then we can add it there.

I think it should be okay to just omit some of the options. (I don't fully understand why exposing `navigationUI` would be harmful though.)

> Good point, maybe the API should act like it's always available even if there's no keyboard. Then I guess we also don't want to reject in `requestFullscreen`.

Agreed.

zcorpan:
So it's something like
```
document.fullscreenOptions.keyboardLock
```
vs
```
document.keyboardLock
```
If we add some other option that should be exposed as a property, that can still be added on `Document`.

> (I don't fully understand why exposing `navigationUI` would be harmful though.)

Is there a use case?

saschanaz:
But it'll be all fullscreen options, why do we need those attributes to be directly in document?

Was also thinking about:

```js
document.fullscreen.keyboardLock
// meaning making fullscreen not a boolean anymore
// less attribute pollution to document
// unless anyone is doing `document.fullscreen == true`....

const resultingOptions = await document.requestFullscreen(options);
resultingOptions.keyboardLock // less hacky way than abusing getter
```

> Is there a use case?

Maybe also feature detection?

zcorpan:
Changing `document.fullscreen` to not return a boolean is likely not web compatible.

```js
const resultingOptions = await document.requestFullscreen(options);
resultingOptions.keyboardLock // less hacky way than abusing getter
```
This means you can only check the state if you called `requestFullscreen`, but that could have happened by other code or the user. Maybe it can be exposed in the `fullscreenchange` event though. Or `navigator.keyboard.locked`

mdevaev:
> It's possible to feature-check that the `keyboardLock` dictionary member is supported without going fullscreen like this: https://software.hixie.ch/utilities/js/live-dom-viewer/saved/14660

It turned out that if the test was failed, then the next legitimate transition to fullscreen mode for the allowed element is blocked (I suppose for security). In other words, I can't test the feature first and then immediately switch to fullscreen mode. But it is suitable for checking whether the keyboardLock options were used or not. I checked for FF151. That is, something better is needed for the initial check. But current solution is enough for my personal case.

In addition, it is not entirely clear which keys should be intercepted and which should not. In FF151, I checked that Ctrl+W and Alt+Tab are not intercepted, although Chromium does it. It's a bit off-top, but nevertheless I would like to mention it: https://bugzilla.mozilla.org/show_bug.cgi?id=2034292

hamishwillee:
Thanks so much @zcorpan et al for your help advice. Hope you can converge this soon.

1. 
   > In addition, it is not entirely clear which keys should be intercepted and which should not. In FF151, I checked that Ctrl+W and Alt+Tab are not intercepted, although Chromium does it. It's a bit off-top, but nevertheless I would like to mention it: https://bugzilla.mozilla.org/show_bug.cgi?id=2034292

   IMO - FWIW the uncertaintly is also a pain for users - it would be great to be a bit more precise about  the interfaces we know can't be intercepted, such as Ctrl+Alt+Delete on Windows. 
   
2. It would be convenient _for me_ to mandate exit via Esc long-press - it simplifies the docs and makes it easier to switch between browsers without having to learn something new.
3. When fullscreen first became a thing I didn't know you could exit it with ESC (I missed the notification - and there were a couple of weeks I was stuck in my browser banging kesy). This lock might have the same problem, at least in the short term.  Worth thinking about multi-tap of esc popping up the notification again?
   EDIT - ignore me, Firefox does this.

zcorpan:
@mdevaev if you want to feature-check and also go fullscreen, instead of throwing you can flip a variable when the `keyboardLock` getter is accessed.

zcorpan:
I think it will be difficult to standardize on which keys are available, since this depends on the platform and different UAs might make different tradeoffs.

It seems we've converged on long-pressing Escape, so the spec can probably have that as a "should".

mdevaev:
> @mdevaev if you want to feature-check and also go fullscreen, instead of throwing you can flip a variable when the `keyboardLock` getter is accessed.

Yea, I've did exactly like this. I'm talking about the situation when we want to test the feature *before* going to fullscreen mode.

fullscreen.bs
```diff
@@ -177,7 +178,28 @@ steps:
 <p class=note>These steps integrate with the <a for=/>event loop</a> defined in HTML. [[!HTML]]
 </div>
 
+<hr>
+
+<p>All <a>elements</a> have an associated <dfn>keyboard lock flag</dfn>. Unless stated otherwise it
+is unset.
+
+<p>A [=document=] has <dfn>keyboard lock</dfn> active when its [=fullscreen element=] is not null
+and the [=fullscreen element=]'s [=keyboard lock flag=] is true, and inactive otherwise.
```
foolip:
```suggestion
and the [=fullscreen element=]'s [=keyboard lock flag=] is set, and inactive otherwise.
```

fullscreen.bs
```diff
@@ -177,7 +178,28 @@ steps:
...
+
+<div algorithm>
+  <p>To <dfn>run the release the keyboard lock steps</dfn> for a <a>document</a>
+  <var>document</var>, run these steps:
+
+  <ol>
+    <li><p>If <var>document</var>'s <a>fullscreen element</a> is null, then return.
 
+    <li><p>If the [=keyboard lock flag=] for <var>document</var>'s <a>fullscreen element</a> is
+    false, then return.
```
foolip:
```suggestion
    <li><p>If <var>document</var>'s <a>fullscreen element</a>'s [=keyboard lock flag=] is unset,
    then return.
```

fullscreen.bs
```diff
@@ -177,7 +178,28 @@ steps:
...
+
+    <li><p>Set the [=keyboard lock flag=] for <var>document</var>'s <a>fullscreen element</a> to
+    false.
```
foolip:
```suggestion
    <li><p>Unset <var>document</var>'s <a>fullscreen element</a>'s [=keyboard lock flag=].
```

fullscreen.bs
```diff
@@ -177,7 +178,28 @@ steps:
...
+
+<p>All <a>elements</a> have an associated <dfn>keyboard lock flag</dfn>. Unless stated otherwise it
+is unset.
+
+<p>A [=document=] has <dfn>keyboard lock</dfn> active when its [=fullscreen element=] is not null
+and the [=fullscreen element=]'s [=keyboard lock flag=] is true, and inactive otherwise.
+
+<div algorithm>
+  <p>To <dfn>run the release the keyboard lock steps</dfn> for a <a>document</a>
+  <var>document</var>, run these steps:
```
foolip: I realize this is modeled after "run the fullscreen steps", but that one's hard to name as an algorithm without the extra "run the" and "steps". In this case:


```suggestion
  <p>To <dfn>release the keyboard lock</dfn> for a <a>document</a> <var>document</var>, run these
  steps:
```

fullscreen.bs
```diff
@@ -506,6 +553,8 @@ could be an open <{dialog}> element.
 
  <li><p>Return <var>promise</var>, and run the remaining steps <a>in parallel</a>.
 
+ <li><p>[=Run the release the keyboard lock steps=] with <var>doc</var>.
```
foolip:
```suggestion
 <li><p>[=Release the keyboard lock=] for <var>doc</var>.
```

fullscreen.bs
```diff
@@ -188,7 +210,13 @@ enum FullscreenNavigationUI {
   "hide"
 };
 
+enum FullscreenKeyboardLock {
+  "browser",
```
foolip, line 199: The naming of this enum seems a bit unusual, although not unprecedented:
- https://w3c.github.io/mediacapture-screen-share/#displaycapturesurfacetype (used instead of "tab" I think, an in contrast to "window")
- https://w3c.github.io/secure-payment-confirmation/#sctn-browser-bound-key-store

Has this already shipped? If so it's too late to change, but if not would "auto" be acceptable?

foolip (commented):
Style nits and just one question about the naming.

fullscreen.bs
```diff
@@ -392,6 +427,17 @@ are:
   <p class=note>The order in which elements are <a lt="fullscreen an element">fullscreened</a>
   is not observable, because <a>run the fullscreen steps</a> is invoked in <a>tree order</a>.
 
+ <li>
+  <p>Depending on the value of <var>options</var>["{{FullscreenOptions/keyboardLock}}"]:
+
+  <dl class=switch>
+   <dt>"{{FullscreenKeyboardLock/browser}}"</dt>
+   <dd><p>Set the [=keyboard lock flag=] for <a>this</a> to true.
```
foolip:
```suggestion
   <dd><p>Se <a>this</a>'s [=keyboard lock flag=].
```

fullscreen.bs
```diff
@@ -392,6 +427,17 @@ are:
...
+
+   <dt>"{{FullscreenKeyboardLock/none}}"</dt>
+   <dd><p>Set the [=keyboard lock flag=] for <a>this</a> to false.
```
foolip:
```suggestion
   <dd><p>Unset <a>this</a>'s [=keyboard lock flag=].
```

fullscreen.bs
```diff
@@ -177,7 +178,27 @@ steps:
...
+
+<p>A <a>document</a> has <dfn>keyboard lock</dfn> active when its <a>fullscreen element</a> is not
+null and the <a>fullscreen element</a>'s <a>keyboard lock flag</a> is set, and inactive otherwise.
+
+<div algorithm>
+  <p>To <dfn>release the keyboard lock</dfn> for a <a>document</a> <var>document</var>, run these
+  steps:
+
+  <ol>
+    <li><p>If <var>document</var>'s <a>fullscreen element</a> is null, then return.
```
foolip: In the case where the fullscreen element is removed, I think the keyboard lock flag will remain set, because of this early return.

Is there any reason to not unconditionally unset the flag?

foolip (commented):
I noticed a problem and am wondering about the more general idea with nested fullscreen scenarios. In both "`requestFullscreen(options)` method steps" and the "exit fullscreen" algorithm, the flag is only set/unset on a single element. But in nested scenarios, there's no guarantee that those elements are the same. Consider this simple case:

```js
elm1.onclick = () => { elm1.requestFullscreen({ keyboardLock: "browser" }); }
elm2.onclick = () => { elm2.requestFullscreen(); }
```

If the elements are clicked in order and later something triggers "fully exit fullscreen", the keyboard lock flag will remain set on _elm1_.

foolip:
How should this interact with the `navigator.keyboard.lock()` API?

zcorpan:
> I noticed a problem and am wondering about the more general idea with nested fullscreen scenarios. In both "`requestFullscreen(options)` method steps" and the "exit fullscreen" algorithm, the flag is only set/unset on a single element. But in nested scenarios, there's no guarantee that those elements are the same. Consider this simple case:
> 
> ```js
> elm1.onclick = () => { elm1.requestFullscreen({ keyboardLock: "browser" }); }
> elm2.onclick = () => { elm2.requestFullscreen(); }
> ```
> 
> If the elements are clicked in order and later something triggers "fully exit fullscreen", the keyboard lock flag will remain set on _elm1_.

No it won't, because fully exit fullscreen calls #unfullscreen-an-element on all fullscreened elements which unsets the keyboard lock flag.

It's true that "exit fullscreen" for elm2 in your example will change the fullscreen element to elm1 and then keyboard lock is active again; this is intentional.

zcorpan:
> How should this interact with the `navigator.keyboard.lock()` API?

For Gecko, the plan is to not support that API.

fullscreen.bs
zcorpan: It has shipped in Safari, in beta in Firefox. (The spec had "application" previously but WebKit implemented "browser".)

fullscreen.bs
zcorpan: Actually I think this algorithm isn't needed at all since "unfullscreen" already unsets the flag. That should handle all cases, right?

zcorpan: Removed now.

foolip:
Can someone prepare the WPT PR so we can merge spec and tests together?

foolip:
@marcoscaceres WDYT about interaction with the keyboard lock API for browsers that support both?

saschanaz:
> Can someone prepare the WPT PR so we can merge spec and tests together?

As listed in the OP, it already exists.

saschanaz:
@foolip does Google have thoughts about usefulness of this API? One of my concerns is that, given browser shortcuts are already locked on fullscreen without keyboard lock on Safari and Chrome, people will not care about this API unless they care about Escape key behavior.

Edit: And the same question also applies to Blink's navigator.keyboard.lock, to me the API looks much less useful now compared to when it was introduced, because of the same reasons above.

foolip:
> > Can someone prepare the WPT PR so we can merge spec and tests together?
> 
> As listed in the OP, it already exists. 

Yes, but "PR for to move to non-tentative TODO"

fullscreen.bs
```diff
@@ -506,7 +538,7 @@ could be an open <{dialog}> element.
 
  <li><p>Return <var>promise</var>, and run the remaining steps <a>in parallel</a>.
 
- <li><p>Run the [=fully unlock the screen orientation steps=] with <var>doc</var>.
+ <li><p>Run the <a>fully unlock the screen orientation steps</a> with <var>doc</var>.
```
saschanaz, line 541: Why is this change?

fullscreen.bs
```diff
@@ -652,6 +689,29 @@ iframe:fullscreen {
...
+have no action while <a>keyboard lock</a> is active, or it could still have the same action but
+allow the web page to call {{Event/preventDefault()}} to cancel the action.
+
+<p>Whenever a <a>document</a>'s <a>keyboard lock</a> is changed from active to inactive, user agents
+must deactivate the keyboard lock and restore the handling of keyboard inputs to the default
+behavior of the user agent and the operating system.
+
+<p>User agents should reserve an additional input for the purposes of exiting fullscreen. There are
+also some system-level key sequences or combinations that cannot be intercepted for security
+reasons, such as Ctrl+Alt+Del on Windows.
```
saschanaz, line 712: Do we want to add implementation note somewhere around here? Something like:

For example, browsers that use Esc key for exiting fullscreen use keyboard lock to prevent immediate exit on key press, and instead require longer press to exit fullscreen. Some of such implementations also implicitly grant partial keyboard lock to fullscreen session, in that case Esc key may be the only key that is affected with explicit keyboard lock. See also https://github.com/whatwg/fullscreen/issues/260.

fullscreen.bs
foolip: Consistency, it was the only [= =] in the spec before this PR, and I converted everything to the old school style

fullscreen.bs
foolip: I've added that with some light edits like browser &rarr; user agent.

@marcoscaceres are you happy with this addition?

foolip: @marcoscaceres I'll go ahead and merge this, but if you want changes later just file a new issue or PR.

foolip (approved):

Issue #133, opened 2018-08-22
Deal with two fullscreen elements in top layer and one element being removed


dtapuska:
With changes in #128 when you have two fullscreen elements in the top layer, and you remove one of them, you would end up making the browser window still in fullscreen, but the document has exited the fullscreen state.

foolip:
This came up again in https://github.com/whatwg/fullscreen/pull/243 and I've written some more tests in https://github.com/web-platform-tests/wpt/pull/50630.

The relevant parts of the spec are step 3.1 and 3.2 of the "removing steps", and step 7.2 of "exit fullscreen".

These are the 4 situations now covered by tests and what I think the spec requires (after https://github.com/whatwg/fullscreen/pull/243):

- remove-first.html: Parent and child elements in fullscreen and the parent is removed. The removing steps visit the parent first, but since it's not the fullscreen element, it just unfullscreens that node. Then it visits the child which is the fullscreen element, and invokes exit fullscreen. There, step 7.2 is reached which unfullscreens that node too. Actually exiting and events follow async.
- remove-last.html: Parent and child elements in fullscreen and the child is removed. Since the child is the fullscreen element, the removing steps invoke exit fullscreen. There, step 7.2 is reached which unfullscreens that node too. Immediately after removal, the parent is the fullscreen element. Then we exit fullscreen async which eventually clears all state.
- remove-first-sibling.html: Two siblings in fullscreen and the first is removed. This should only unfullscreen that element and not exit fullscreen or fire any events.
- remove-last-sibling.html: Two siblings in fullscreen and the second is removed. The second element is the fullscreen element, and this is much like the remove-last.html case.

AFAICT we don't get any dangling state, but the behavior is a bit curious. AFAICT there isn't any reason we have to exit fullscreen if there are other fullscreen elements still in top layer. The special case of the topmost such element seems unnecessary, but probably also harmless.

cc @nt1m

PR #243, opened 2025-02-04
Fix dangling fullscreen flag when exiting a disconnected element

### LINKED ISSUE #217: Removing steps on fullscreen element leave a dangling fullscreen flag
If you have a fullscreen element, then move it to a different parent, it triggers the removing steps which does at step 3:
1. If node is document’s [fullscreen element](https://fullscreen.spec.whatwg.org/#fullscreen-element), [exit fullscreen](https://fullscreen.spec.whatwg.org/#exit-fullscreen) document.
2. Otherwise...
3. If document’s [top layer](https://fullscreen.spec.whatwg.org/#top-layer) [contains](https://infra.spec.whatwg.org/#list-contain) node, [remove](https://infra.spec.whatwg.org/#list-remove) node from document’s [top layer](https://fullscreen.spec.whatwg.org/#top-layer).

But in the [exit fullscreen algorithm](https://fullscreen.spec.whatwg.org/#exit-fullscreen), the fullscreen flag is only removed at step 14.3. This is problematic, because that step runs in parallel, so step 3.3 of the removing steps which removes the fullscreen element from the top layer will run before.

That means we end up stopping the exit fullscreen at step 11, so we end up never removing the fullscreen flag, which is problematic if the removed element is re-appended in the document.

Chromium handles it specifically at step 7. of exit fullscreen, by invoking the "unfullscreen doc's fullscreen element" as a step 7.2.

nt1m:
Fixes #217

   * WebKit
   * Chromium? (seems like shipping behavior)
   * Gecko? (seems like shipping behavior)
   * wpt.fyi/fullscreen/model/move-fullscreen-element.html
   * https://github.com/web-platform-tests/wpt/pull/50625
   * Chromium: N/A
   * Gecko: N/A
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=253121
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: N/A
- [ ] The top of this comment includes a [clear commit message](https://github.com/whatwg/meta/blob/main/COMMITTING.md) to use. 

***
<a href="https://whatpr.org/fullscreen/243.html" title="Last updated on Feb 11, 2025, 12:20 PM UTC (02b02e5)">Preview</a> | <a href="https://whatpr.org/fullscreen/243/3cd33da...02b02e5.html" title="Last updated on Feb 11, 2025, 12:20 PM UTC (02b02e5)">Diff</a>

### DIFF
--- a/fullscreen.bs
+++ b/fullscreen.bs
@@ -495,9 +495,11 @@ could be an open <{dialog}> element.
 
  <li><p>If <var>doc</var>'s <a>fullscreen element</a> is not <a>connected</a>:
   <ol>
-   <li><p><a for=set>Append</a> ({{fullscreenchange}}, <var>doc</var>'s
-   <a>fullscreen element</a>) to <var>doc</var>'s
-   <a>list of pending fullscreen events</a>.
+   <li><p><a for=set>Append</a> ({{fullscreenchange}}, <var>doc</var>'s <a>fullscreen element</a>)
+   to <var>doc</var>'s <a>list of pending fullscreen events</a>.
+
+   <li><p><a lt="unfullscreen an element">Unfullscreen</a> <var>doc</var>'s <a>fullscreen
+   element</a>.
   </ol>
 
  <li><p>Return <var>promise</var>, and run the remaining steps <a>in parallel</a>.


foolip (approved):
Per https://github.com/whatwg/fullscreen/issues/217#issuecomment-2647289713 I think this is the right fix. I've sent https://github.com/whatwg/fullscreen/pull/244 to fix Bikeshed errors so that this will build cleanly.

Also note that there is a remaining question mark in https://github.com/whatwg/fullscreen/issues/133 that was filed while this very step was discussed in https://github.com/whatwg/fullscreen/pull/128.

foolip:
I have reviewed https://github.com/web-platform-tests/wpt/blob/master/fullscreen/model/move-fullscreen-element.html and agree it's an appropriate test for this change. https://wpt.fyi/results/fullscreen/model/move-fullscreen-element.html?label=experimental&label=master&aligned shows it passing in Chrome/Edge/Firefox already.

However, I think the reason the test passes in Firefox is different. If the comment [here](https://searchfox.org/mozilla-central/rev/206eaea9a2fd4307da16e1614cd934920368165a/dom/base/Element.cpp#2319-2320) is correct, Gecko _fully_ exits fullscreen when the fullscreen element is removed.

The difference should be observable in a test written for https://github.com/whatwg/fullscreen/issues/133.

nt1m:
> Per [#217 (comment)](https://github.com/whatwg/fullscreen/issues/217#issuecomment-2647289713) I think this is the right fix. I've sent #244 to fix Bikeshed errors so that this will build cleanly.

Thanks! I appreciate it.

> Also note that there is a remaining question mark in #133 that was filed while this very step was discussed in #128.

Hmm, I guess I don't feel too strongly. I just went with whatever Chrome shipped to fix the YouTube regression. Gecko's approach also works too.

> I have reviewed https://github.com/web-platform-tests/wpt/blob/master/fullscreen/model/move-fullscreen-element.html and agree it's an appropriate test for this change. https://wpt.fyi/results/fullscreen/model/move-fullscreen-element.html?label=experimental&label=master&aligned shows it passing in Chrome/Edge/Firefox already.
> 
> However, I think the reason the test passes in Firefox is different. If the comment [here](https://searchfox.org/mozilla-central/rev/206eaea9a2fd4307da16e1614cd934920368165a/dom/base/Element.cpp#2319-2320) is correct, Gecko _fully_ exits fullscreen when the fullscreen element is removed.
> 
> The difference should be observable in a test written for #133.

I don't currently have time to do this, but I assume we could make a variant of the test that pushes 2 things in fullscreen, then moves one of them to see the difference?

foolip:
> I don't currently have time to do this, but I assume we could make a variant of the test that pushes 2 things in fullscreen, then moves one of them to see the difference?

I started writing such a test, but found that we already have it tested for removing both positions in the stack in a nested fullscreen case:

https://wpt.fyi/results/fullscreen/model/remove-first.html?label=master&label=experimental&aligned
https://wpt.fyi/results/fullscreen/model/remove-last.html?label=master&label=experimental&aligned

@nt1m the failures in Safari seem unrelated to the problem at hand. Is there some tweak to the tests that would make them run correctly and reveal what the current consensus behavior is?

nt1m:
> > I don't currently have time to do this, but I assume we could make a variant of the test that pushes 2 things in fullscreen, then moves one of them to see the difference?
> 
> I started writing such a test, but found that we already have it tested for removing both positions in the stack in a nested fullscreen case:
> 
> https://wpt.fyi/results/fullscreen/model/remove-first.html?label=master&label=experimental&aligned https://wpt.fyi/results/fullscreen/model/remove-last.html?label=master&label=experimental&aligned
> 
> @nt1m the failures in Safari seem unrelated to the problem at hand. Is there some tweak to the tests that would make them run correctly and reveal what the current consensus behavior is?

I think I fixed most of the WPT failures recently, but it hasn't reached STP.

Might be more useful to check these results: https://searchfox.org/wubkat/search?q=imported%2Fw3c%2Fweb-platform-tests%2Ffullscreen%2Fmodel%2Fremove-&path=&case=false&regexp=false

foolip:
Thanks @nt1m! I've looked at those results, and also run the tests manually in different browsers.

Chrome passes both remove-first.html and remove-last.html.

Firefox passes remove-first.html but fails remove-last.html. The failures is on [this assert](https://github.com/web-platform-tests/wpt/blob/2897a7bbe05d9b3e43deb694c7ffb5eb33d8f7a7/fullscreen/model/remove-last.html#L39), because `document.fullscreenElement` was synchronously changed to null instead of the first element being left as the fullscreen element. I assume this is because of the synchronous "fully exit fullscreen" call linked in https://github.com/whatwg/fullscreen/pull/243#issuecomment-2647847263.

Safari stable fails both tests, but per the results @nt1m shared the remove-first.html failures has been fixed. I can't tell from the failure message which exact assert failed. I've sent https://github.com/web-platform-tests/wpt/pull/50625 to make the tests more debuggable.

foolip:
Since this PR fixes a very real problem and matches an existing implementation (Chromium) I'll go ahead and merge it now. In https://github.com/whatwg/fullscreen/issues/133 I'll summarize the current situation and things we might still change on top of this PR.

PR #255, opened 2026-03-10
Bikeshed has migrated to new URL

theIDinside:
And beyond that, we also now have to provide the -F type=bikeshed-spec arguments

See https://github.com/w3c/spec-generator/wiki/Migrating-Bikeshed-HTTP-API-usage-to-spec%E2%80%90generator


annevk:
We have to make this change in https://github.com/whatwg/spec-factory and then apply it everywhere.

annevk:
#256

